PR #19, opened 2015-02-10
154 - Adjusts sphinx theme to use the RTD theme if available

bmbouter:
https://pulp.plan.io/issues/154

The new implementation only specifies a html_static_path if the `sphinx_rtd_theme` is available. I don't see any `_static` content in the docs anyway so this shouldn't matter. The docs build normally for me with no CSS errors.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -11,7 +11,10 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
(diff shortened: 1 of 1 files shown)


rbarlow:
LGTM. The Travis failure is due to platform adding dependencies without putting them in its setup.py files, so don't worry about it.

PR #20, opened 2015-02-17
Add support for syncing packages from PyPI.

rbarlow:
closes #104

https://pulp.plan.io/issues/104

This PR must be merged, and the Travis file must be updated not to use my pulp branch first:

https://github.com/pulp/pulp/pull/1662

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,9 @@ before_install:
   - "popd"
   - "popd"
 install:
+  # This is needed to build M2Crypto
+  - "sudo apt-get install swig"
   - "pip install -r test_requirements.txt"
-  - "pip install -r requirements.txt"
   - "pushd .."
   - "python nectar/setup.py develop"
   - "pulp/manage_setup_pys.sh develop"
--- a/extensions_admin/test/unit/admin/test_cudl.py
+++ b/extensions_admin/test/unit/admin/test_cudl.py
@@ -1,6 +1,9 @@
+"""
+This module contains tests for the pulp_python.extensions.admin.cudl module.
+"""
 import unittest
 
-from mock import Mock
+import mock
 from pulp.common.constants import REPO_NOTE_TYPE_KEY
 from pulp.common.plugins.importer_constants import KEY_FEED
 from pulp.devel.unit.util import compare_dict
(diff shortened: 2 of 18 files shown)


beav:
@rbarlow the `.travis.yml` probably needs to be updated, its pointed at 2.5-release for platform.

From travis run:
```
ImportError: cannot import name doc_check
```

rbarlow:
I made this PR to address this Travis issue:

https://github.com/pulp/pulp/pull/1662

rbarlow:
The above Travis run passed with a modified Travis config that build from my PR in #1662. Once we have that other PR merged, I'll redo this Travis config to use Pulp's master again so we can be sure it passes that way. I.e., this Travis config is not the way I want it to be for merging this PR.

.travis.yml
```diff
@@ -5,15 +5,22 @@ python:
   - "pypy"
 before_install:
   - "pushd .."
-  - "git clone https://github.com/pulp/pulp.git --branch 2.5-release"
+  - "echo 'Remove this'"
```
asmacdo: Was this a note to yourself?

.travis.yml
```diff
@@ -5,15 +5,22 @@ python:
...
+  - "rm -rf pulp"
+  - "git clone https://github.com/pulp/pulp.git --branch master"
   - "git clone https://github.com/pulp/nectar.git"
   - "pushd nectar"
   - "git checkout python-nectar-1.1.6-1"
   - "popd"
   - "popd"
 install:
+  # This is needed to build M2Crypto
+  - "sudo apt-get install swig"
```
asmacdo: This is here twice.

extensions_admin/test/unit/admin/test_cudl.py
```diff
@@ -1,15 +1,33 @@
+"""
+This module contains tests for the pulp_python.extensions.admin.cudl module.
+"""
 import unittest
 
-from mock import Mock
 from pulp.common.constants import REPO_NOTE_TYPE_KEY
 from pulp.common.plugins.importer_constants import KEY_FEED
 from pulp.devel.unit.util import compare_dict
+import mock
```
asmacdo: 3rd party import.

plugins/pulp_python/plugins/importers/sync.py
```diff
@@ -0,0 +1,267 @@
+"""
+This module contains the necessary means for a necessary means for syncing packages from PyPI.
```
asmacdo: s/a necessary means for//

.travis.yml
rbarlow: On 02/20/2015 05:44 PM, Austin wrote:
> Was this a note to yourself?

Yeah, the lines below this are just temporary so that I can test the
changes I made in platform. I don't want this or the lines below it to
actually be merged, but I wanted to make sure what I was doing in that
other PR made this one pass in Travis. Once that PR is merged, I will
remove these lines before I merge. The PR description has something
along those lines in it as well.

.travis.yml
rbarlow: On 02/20/2015 05:45 PM, Austin wrote:
> This is here twice.

This is also on purpose. I needed it to be in that block above that will
get removed, so I wanted to make sure it was here too so that it doesn't
get removed when I fix the above block.

plugins/pulp_python/plugins/importers/sync.py
rbarlow: On 02/20/2015 06:08 PM, Austin wrote:
>> +This module contains the necessary means for a necessary means for syncing packages from PyPI.
> 
> s/a necessary means for//

Dude, haven't you seen The Big Lebowski?

beav:
LGTM :snake: 

one question: is it expected that autopublish on sync is disabled? It may be worth documenting that autopublish is not enabled by default.

rbarlow:
On 02/25/2015 10:59 AM, Chris Duryee wrote:
> one question: is it expected that autopublish on sync is disabled? It
> may be worth documenting that autopublish is not enabled by default.

Hmmm, I hadn't payed attention to that. I'm not sure. Some of our
importers do it one way and some the other. It isn't on purpose that it
does it this way. Do you think it should? What do others think?

beav:
Personally I think it's OK to not autopublish since it's a new importer :koala:. Maybe just add a note saying something like "if you want autopublish, add `--auto-publish` to the repo create command"?

Also, the help text is a little off if you just run `pulp-admin python repo create`:

```
  --auto-publish - if "true", on each successful sync the repository will
                   automatically be published; if "false" content will only be
                   available after manually publishing the repository; defaults
                   to "true"
```

rbarlow:
I decided to update the help text to say that the default is false, rather than going through the trouble of making it true. We can change it to true in a separate PR if that is desired.

PR #22, opened 2015-02-23
add dist file with supported distributions for pulp_python

barnabycourt:


rbarlow:
Travis probably won't pass this build until https://github.com/pulp/pulp_python/pull/20 is reviewed and merged.

rbarlow:
As discussed via e-mail, this travis failure is due to a fix that went to master instead of to 0.0-testing.

If we need to release another 0.0 version, I suppose we'll need the work you are doing in this PR. I don't believe we will do that, so if you want to, you could close this and re-submit against master.

If you do think we want this in 0.0-testing, let me know and I'll review it here and we can just ignore Travis.

beav:
Tests are green now :green_heart: 

LGTM

PR #23, opened 2015-02-25
add import order checks to flake8

beav:
Note that local packages need to be specified to flake8 via run-tests.py


beav:
this requires https://github.com/pulp/pulp/pull/1681 for travis to pass

jortel:
LGTM

PR #26, opened 2015-03-22
Remove the doc and coverage checks from run-tests.py.

rbarlow:
doc_check and coverage checking are new in Pulp's master branch, but we want 1.0.0 of the
plugins to work against Pulp 2.5 ideally, or at least 2.6.0. This commit removes the doc_check
and coverage check from run-tests.

Also, 2.5 did not list its dependencies in its setup.py, so this adds them to the .travis.yml
file.

This change must not be propagated to master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ python:
   - "pypy"
 before_install:
   - "pushd .."
-  - "git clone https://github.com/pulp/pulp.git --branch master"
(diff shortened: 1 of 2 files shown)


beav:
LGTM

rbarlow:
I just confirmed that this was merged to master with the ours strategy, as it should have been. I believe this will allow the 1.0-* branches to have this fix (which is necessary), but also that merging forward will not introduce this change to master (which is undesirable).  Thanks for the review!

PR #27, opened 2015-03-22
Set the version to 1.0.0-0.1.alpha.

rbarlow:


dkliban:
LGTM

extensions_admin/setup.py
```diff
@@ -3,7 +3,7 @@
 
 setup(
     name='pulp_python_extensions_admin',
-    version='0.0.0',
+    version='1.0.0',
```
beav: This is slightly different than platform, which would be something like 1.0.0a1 I believe. I think both ways work OK though :koala:

rbarlow: I don't personally think it's necessary for the setup.py files to have this, but if we have a team policy that it should I can change it later. For now, I'll leave it as is since primarily the spec file will communicate its "alphaness".

PR #36, opened 2015-04-08
Update the .travis.yml & run-tests to maintain Pulp 2.6 Compatibility

barnabycourt:


run-tests.py
```diff
@@ -5,7 +5,7 @@
 import subprocess
 import sys
 
-from pulp.devel import doc_check
+# from pulp.devel import doc_check
```
bmbouter: Can these be removed entirely? I don't think checking in commented code is a good practice. We can use version control to restore the statements easily.

bmbouter:
Rebasing would be good. LGTM.

PR #37, opened 2015-04-13
Rename .rst

ryanhiebert:
It's already an rST file, so it's just changing the extension.


ryanhiebert:
The build doesn't pass, but any fix to make it pass would not be related to this change.

beav:
@ryanhiebert Thanks for your patch! I just merged #38 which should fix this test. Can you rebase and update the PR?

ryanhiebert:
done.

beav:
LGTM, merging patch.

Thanks again! :sunglasses:

PR #40, opened 2015-05-06
DOS Line Ending Support/Use the PKG-INFO file with the shortest path

vitaminmoo:
Support PKG-INFO files with DOS line endings.

Some upstream packages have DOS line endings. Previously this caused pulp to import them with corrupted names and versions.

Use the PKG-INFO file with the shortest path
    
Without this packages that contain test PKG-INFO files that are completely different than the primary one may fail depending on the order of the PKG-INFO files in the tar file. An example of this is the 'nose' package.

I've updated test coverage, but I may have gotten a little lost, so please pay close attention to the changes in that area.

### DIFF
--- a/plugins/pulp_python/plugins/models.py
+++ b/plugins/pulp_python/plugins/models.py
@@ -57,8 +57,9 @@ class Package(object):
 
             # Build a list of tuples of all the attributes found in the metadata. Ignore attributes
             # with a leading underscore, as they are not part of the metadata.
+            # rstrip for DOS line ending support
             try:
-                required_attrs = [attr for attr in cls._ATTRS if attr[0] != '_']
+                required_attrs = [attr.rstrip() for attr in cls._ATTRS if attr[0] != '_']
                 attrs = dict()
                 for attr in required_attrs:
                     attrs[attr] = re.search('^%s: (?P<field>.*)$' % cls._metadata_label(attr),
--- a/plugins/pulp_python/plugins/models.py
+++ b/plugins/pulp_python/plugins/models.py
@@ -18,7 +18,8 @@ class Package(object):
     # The full list of supported attributes. Attributes beginning with underscore are specific to
     # this module and are not found in PKG-INFO.
     _ATTRS = ('name', 'version', 'summary', 'home_page', 'author', 'author_email', 'license',
-              'description', 'platform', '_filename', '_checksum', '_checksum_type')
+              'description', 'platform', '_filename', '_checksum', '_checksum_type',
+              '_metadata_file')
 
     @classmethod
     def from_archive(cls, archive_path):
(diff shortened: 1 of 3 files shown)


rbarlow:
My IRC client started majorly lagging, so I'm not sure if you saw these messages, so I'll repeat here:

```
18:01:24   vitaminmoo | Hmm, re: pull/40, pr tester says travisci failed, but the link
                      | to details renders blank    
18:03:49   bowlofeggs | vitaminmoo: hmmm, this is surprising, but it looks like the   
                      | test environment is just missing kombu:    
                      | https://travis-ci.org/pulp/pulp_python/jobs/61533597    
18:04:00   bowlofeggs | vitaminmoo: that's probably Not Your Fault™    
18:04:14   bowlofeggs | vitaminmoo: i suspect we've got a problem in the pulp setup.py
18:04:33   bowlofeggs | pulp's setup.py should be installing all it's dependencies and 
                      | travis should be getting them that way    
18:04:43   bowlofeggs | vitaminmoo: i advise you not to worry about it ☺    
18:05:03   bowlofeggs | vitaminmoo: i need to leave for the day, but i did most of the 
                      | python stuff so i'll look at it real soon    
18:05:12   bowlofeggs | from the description, it does sound important!    
18:05:38   bowlofeggs | thanks so much for the PR, and i'll get to it as soon as i can 
18:05:41   bowlofeggs | vitaminmoo++
```

barnabycourt:
ok test

rbarlow:
@barnabycourt the Jenkins failure is due to a missing test dependency that can currently only be installed via pip. i use virtualenvs to do that typically.

rbarlow:
I dug deeper in the Travis output, and it looks like Django 1.8 fails to pip install in Python 2.6, which then causes the rest of the pip install to exit (thus never installing Kombu).

This means we have the option of having Pulp depend on 1.4 <= Django < 1.8, or we need to stop having Travis test Python 2.6 with the Python plugins.

Either way, none of these problems are due to this pull request so no need for you to worry about them. I'll think about which path is best and I'll rerun the tests after fixing that.

rbarlow:
This PR will fix the Jenkins issue here:

https://github.com/pulp/pulp_packaging/pull/59

rbarlow:
This PR will fix the Travis issue:

https://github.com/pulp/pulp/pull/1862

rbarlow:
@vitaminmoo The code changes look good, and you maintained 100% test coverage!

plugins/pulp_python/plugins/models.py
```diff
@@ -18,7 +18,8 @@ class Package(object):
     # The full list of supported attributes. Attributes beginning with underscore are specific to
     # this module and are not found in PKG-INFO.
     _ATTRS = ('name', 'version', 'summary', 'home_page', 'author', 'author_email', 'license',
-              'description', 'platform', '_filename', '_checksum', '_checksum_type')
+              'description', 'platform', '_filename', '_checksum', '_checksum_type',
+              '_metadata_file')
```
rbarlow: I'm curious - why store the metadata file here? It's not harming anything, but I am wondering what use case it is needed for.

rbarlow:
I'm working on getting a lot of PRs merged to make these CI tests pass. It may take another day or two, so sit tight. Thanks!

rbarlow:
ok test

rbarlow:
ok test

rbarlow:
ok test

rbarlow:
Thanks for the contribution!

rbarlow:
I almost forgot - we have an AUTHORS file. If you want your name in there, feel free to make a PR against it. Or I can just add your name/email from your commit message for you if you prefer.

plugins/pulp_python/plugins/models.py
vitaminmoo: Late reply - Just seemed like the most reasonable existing place to put it.

rbarlow:
Hello @vitaminmoo! Today I cherry-picked these two commits into our 1.0-testing branch:

https://github.com/pulp/pulp_python/commits/1.0-testing

This will become a 1.0.1 release, and I'm planning to release that with the upcoming Pulp 2.7. It hasn't been built yet, but I think it should land in the next Pulp 2.7 beta. Thanks again for your contribution!

rbarlow:
Hello again @vitaminmoo! I have a good surprise - your fixes actually got included in the 2.6.3 beta repository. Check it out and let me know how it goes. You should see pulp-python-plugins-1.0.1-0.0.beta.fc21.noarch.rpm in the various arch beta repos. For example, RHEL 7's 2.6 beta repo is here:

https://repos.fedorapeople.org/repos/pulp/pulp/beta/2.6/7Server/x86_64/

PR #42, opened 2015-05-26
Add nosexcover to test_requirements.txt.

rbarlow:


rbarlow:
The Travis failure is strange, but I believe it is not related to this PR.

PR #46, opened 2015-06-18
Added fc22 to dist_list.txt

dkliban:


ipanova:
ok test

rbarlow:
ok test

dist_list.txt
```diff
@@ -1 +1 @@
-el6 el7 fc20 fc21
\ No newline at end of file
+el6 el7 fc20 fc21 fc22
```
rbarlow: I think we should also drop support for F20, since it is only supported for two more days.

rbarlow:
The tests fail on the 1.0-* branches due to some nonsense that isn't worth fixing IMO. So don't worry about that. LGTM!

PR #51, opened 2015-10-14
repo.working_dir can be None

vjancik:
I encountered an error where if you delete a repository after adding something to it but before publishing it, repo.working_dir is None, which causes an error in the distributor_removed routine.

Traceback:
https://paste.fedoraproject.org/279066/82722614/
Reproduced:
http://fpaste.org/279071/27799144/
https://pulp.plan.io/issues/1349

This was fixed in pulp_ostree in an identical method:
https://pulp.plan.io/issues/1096
https://github.com/pulp/pulp_ostree/commit/49274ee493dff222ffcc82985238abfa6e25d220

### DIFF
--- a/plugins/test/unit/plugins/distributors/test_web.py
+++ b/plugins/test/unit/plugins/distributors/test_web.py
@@ -56,6 +56,21 @@ class TestBasics(unittest.TestCase):
 
         self.assertEquals(0, len(os.listdir(self.working_dir)))
 
+    @patch('pulp_python.plugins.distributors.web.configuration.get_master_publish_dir')
+    @patch('pulp_python.plugins.distributors.web.configuration.get_web_publish_dir')
+    def test_distributor_removed_dir_is_none(self, mock_web, mock_master):
+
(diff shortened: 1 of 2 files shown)


plugins/test/unit/plugins/distributors/test_web.py
```diff
@@ -56,6 +56,21 @@ def test_distributor_removed(self, mock_web, mock_master):
 
         self.assertEquals(0, len(os.listdir(self.working_dir)))
 
+    @patch('pulp_ostree.plugins.distributors.web.configuration.get_master_publish_dir')
```
ipanova: please update pulp_ostree to pulp_python

ipanova:
@vjancik I reported the issue https://pulp.plan.io/issues/1349 so can you please squash the commits and add the commit message like described here http://pulp.readthedocs.org/en/latest/dev-guide/contributing/branching.html?highlight=commit#commit-messages

thank you!

vjancik:
Squashed the commits, modified the commit message.

PR #56, opened 2016-01-03
converted to use mongoengine

mhrivnak:
converted to use mongoengine

fixes #877
fixes #1465

https://pulp.plan.io/issues/877
https://pulp.plan.io/issues/1465

1465 was fixed by accident in the course of doing the conversion

### DIFF
--- a/plugins/pulp_python/plugins/distributors/steps.py
+++ b/plugins/pulp_python/plugins/distributors/steps.py
@@ -1,11 +1,14 @@
 from gettext import gettext as _
+import itertools
 import logging
 import os
 from xml.etree import cElementTree as ElementTree
 
 from pulp.plugins.util.publish_step import AtomicDirectoryPublishStep, PluginStep
+from pulp.server.controllers import repository as repo_controller
 
 from pulp_python.common import constants
+from pulp_python.plugins import models
 from pulp_python.plugins.distributors import configuration
 
 
@@ -30,7 +33,7 @@ class PublishContentStep(PluginStep):
         """
         Publish all the python files themselves by creating the symlinks to the storage paths.
         """
-        for name, packages in _get_packages(self.get_conduit()).items():
+        for name, packages in _get_packages(self.get_conduit().repo_id).items():
             for package in packages:
                 relative_path = _get_package_path(name, package['filename'])
                 symlink_path = os.path.join(self.parent.web_working_dir, relative_path)
@@ -62,7 +65,7 @@ class PublishMetadataStep(PluginStep):
         os.makedirs(simple_path)
         simple_index_path = os.path.join(simple_path, 'index.html')
 
-        packages = _get_packages(self.get_conduit())
+        packages = _get_packages(self.get_conduit().repo_id)
 
         with open(simple_index_path, 'w') as index:
             html = ElementTree.Element('html')
@@ -168,27 +171,27 @@ def _get_package_path(name, filename):
     return os.path.join('packages', 'source', name[0], name, filename)
 
 
-def _get_packages(conduit):
+def _get_packages(repo_id):
     """
     Build and return a data structure of the available packages. The keys each index a list of
     dictionaries. The inner dictionaries are of the form
     {'version': VERSION, 'filename': FILENAME, 'checksum': MD5SUM, 'checksum_type': TYPE,
      'storage_path': PATH}
 
-    :param conduit: A SingleRepoUnitsMixin object configured for the repo you are publishing.
-    :type  conduit: pulp.plugins.conduits.mixins.SingleRepoUnitsMixin
+    :param repo_id: ID of the repo being published.
+    :type  repo_id: basestring
     :return:        A dictionary of all the packages in the repo to be published
     :rtype:         dict
     """
     packages = {}
-    for p in conduit.get_units():
-        name = p.unit_key['name']
-        if name not in packages:
-            packages[name] = []
-        packages[name].append(
-            {'version': p.unit_key['version'],
-             'filename': p.metadata['_filename'],
-             'checksum': p.metadata['_checksum'],
-             'checksum_type': p.metadata['_checksum_type'],
+    fields = ('version', '_filename', '_checksum', '_checksum_type', 'name', '_storage_path')
+    unit_querysets = repo_controller.get_unit_model_querysets(repo_id, models.Package)
+    unit_querysets = (q.only(*fields) for q in unit_querysets)
+    for p in itertools.chain(*unit_querysets):
+        packages.setdefault(p.name, []).append(
+            {'version': p.version,
+             'filename': p._filename,
+             'checksum': p._checksum,
+             'checksum_type': p._checksum_type,
              'storage_path': p.storage_path})
     return packages
--- a/plugins/pulp_python/plugins/importers/importer.py
+++ b/plugins/pulp_python/plugins/importers/importer.py
@@ -1,9 +1,10 @@
 from gettext import gettext as _
+from itertools import chain
 import shutil
 import tempfile
 
 from pulp.plugins.importer import Importer
-from pulp.server.db.model import criteria
+from pulp.server.controllers import repository as repo_controller
 
 from pulp_python.common import constants
 from pulp_python.plugins import models
@@ -65,11 +66,12 @@ class PythonImporter(Importer):
         :rtype:                list
         """
         if units is None:
-            search = criteria.UnitAssociationCriteria(type_ids=[constants.PACKAGE_TYPE_ID])
-            units = import_conduit.get_source_units(criteria=search)
+            units = chain(*repo_controller.get_unit_model_querysets(source_repo.repo_obj.repo_id,
+                                                                    models.Package))
 
+        units = list(units)
         for u in units:
-            import_conduit.associate_unit(u)
+            repo_controller.associate_single_unit(dest_repo.repo_obj, u)
 
         return units
 
@@ -181,11 +183,9 @@ class PythonImporter(Importer):
         :rtype:           dict
         """
         package = models.Package.from_archive(file_path)
-        package.init_unit(conduit)
-
-        shutil.move(file_path, package.storage_path)
-
-        package.save_unit(conduit)
+        package.save()
+        package.import_content(file_path)
+        repo_controller.associate_single_unit(repo.repo_obj, package)
 
         return {'success_flag': True, 'summary': {}, 'details': {}}
 
--- a/plugins/pulp_python/plugins/importers/sync.py
+++ b/plugins/pulp_python/plugins/importers/sync.py
@@ -1,18 +1,20 @@
 """
 This module contains the necessary means for a necessary means for syncing packages from PyPI.
 """
+import contextlib
 from cStringIO import StringIO
 from gettext import gettext as _
 import json
 import logging
 import os
-import shutil
 from urlparse import urljoin
 
+import mongoengine
 from nectar import request
 from pulp.common.plugins import importer_constants
 from pulp.plugins.util import publish_step
-from pulp.server.db.model import criteria
+from pulp.plugins.util.publish_step import GetLocalUnitsStep
+from pulp.server.controllers import repository as repo_controller
 
 from pulp_python.common import constants
 from pulp_python.plugins import models
@@ -23,10 +25,17 @@ _logger = logging.getLogger(__name__)
 
 class DownloadMetadataStep(publish_step.DownloadStep):
     """
-    This DownloadStep subclass contains the code to process the downloaded manifests and decide what
-    to download from the feed. It does this as it gets each metadata file to spread the load on the
-    database.
+    This DownloadStep subclass contains the code to process the downloaded manifests and determine
+    what is available from the feed. It does this as it gets each metadata file to spread the load
+    on the database.
     """
+
+    def __init__(self, *args, **kwargs):
+        """
+        Injects download requests into the call to the parent class's __init__
+        """
+        kwargs['downloads'] = self.generate_download_requests()
+        super(DownloadMetadataStep, self).__init__(*args, **kwargs)
 
     def download_failed(self, report):
         """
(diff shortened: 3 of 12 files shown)


plugins/pulp_python/plugins/distributors/steps.py
```diff
@@ -168,27 +171,27 @@ def _get_package_path(name, filename):
...
-        if name not in packages:
-            packages[name] = []
-        packages[name].append(
-            {'version': p.unit_key['version'],
-             'filename': p.metadata['_filename'],
-             'checksum': p.metadata['_checksum'],
-             'checksum_type': p.metadata['_checksum_type'],
+    fields = ('version', '_filename', '_checksum', '_checksum_type', 'name', '_storage_path')
+    unit_querysets = repo_controller.get_unit_model_querysets(repo_id, models.Package)
+    unit_querysets = (q.only(*fields) for q in unit_querysets)
```
bmbouter: :+1:

plugins/pulp_python/plugins/importers/importer.py
```diff
@@ -65,11 +66,12 @@ def import_units(self, source_repo, dest_repo, import_conduit, config, units=Non
...
         if units is None:
-            search = criteria.UnitAssociationCriteria(type_ids=[constants.PACKAGE_TYPE_ID])
-            units = import_conduit.get_source_units(criteria=search)
+            units = chain(*repo_controller.get_unit_model_querysets(source_repo.repo_obj.repo_id,
+                                                                    models.Package))
 
+        units = list(units)
         for u in units:
-            import_conduit.associate_unit(u)
+            repo_controller.associate_single_unit(dest_repo.repo_obj, u)
```
bmbouter: Where is the call to `rebuild_content_unit_counts()` for this single unit association? The old style `import_conduit.associate_unit()` does update the repo content unit count but with the mongoengine based methods `rebuild_content_unit_counts()` needs to be explicitly called.

bmbouter: Calling it outside the loop after a bunch of them are associated allows all imported units to be handled with 1 call.

plugins/pulp_python/plugins/importers/importer.py
```diff
@@ -181,11 +183,10 @@ def upload_unit(self, repo, type_id, unit_key, metadata, file_path, conduit, con
...
         package = models.Package.from_archive(file_path)
-        package.init_unit(conduit)
-
-        shutil.move(file_path, package.storage_path)
-
-        package.save_unit(conduit)
+        package.save()
+        package.import_content(file_path)
+        repo_controller.associate_single_unit(repo.repo_obj, package)
+        repo_controller.rebuild_content_unit_counts(repo.repo_obj)
```
bmbouter: Since you're only adding a single unit, you could replace this with the more explicit/efficient:

`repo_controller.update_unit_count(repo.repo_obj, package._content_type_id, 1)`

plugins/pulp_python/plugins/importers/sync.py
```diff
@@ -1,18 +1,20 @@
 """
 This module contains the necessary means for a necessary means for syncing packages from PyPI.
 """
+import contextlib
```
bmbouter: I didn't know about this; nice.

plugins/pulp_python/plugins/models.py
```diff
@@ -3,23 +3,58 @@
...
+    author = StringField()
+    author_email = StringField()
+    description = StringField()
+    home_page = StringField()
+    license = StringField()
+    platform = StringField()
+    summary = StringField()
+
+    _checksum = StringField()
+    _checksum_type = StringField(default=DEFAULT_CHECKSUM_TYPE)
```
bmbouter: From code in the PR at least, it looks like all instantiations of `Package` specify `_checksum_type` to `md5`. Should that just be the default?

plugins/pulp_python/plugins/models.py
```diff
@@ -3,23 +3,58 @@
...
+
+    author = StringField()
+    author_email = StringField()
+    description = StringField()
+    home_page = StringField()
+    license = StringField()
+    platform = StringField()
+    summary = StringField()
+
+    _checksum = StringField()
```
bmbouter: What about a pre-save signal to calculate `_checksum` if it is not set.

plugins/pulp_python/plugins/importers/importer.py
mhrivnak: The platform calls it for this operation. I think the platform should probably be responsible for calling it for all operations, but that's a mess we haven't cleaned up yet.

https://github.com/pulp/pulp/blob/ed36708132e18bac4ab8b58d0b56cdb2ffe95685/server/pulp/server/managers/repo/unit_association.py#L259

plugins/pulp_python/plugins/models.py
bmbouter: We did this in pulp_puppet and it was easily added to the `import_content` implementation.

https://github.com/pulp/pulp_puppet/blob/master/pulp_puppet_plugins/pulp_puppet/plugins/db/models.py#L141-L143

plugins/pulp_python/plugins/models.py
```diff
@@ -3,23 +3,58 @@
...
 
-class Package(object):
+# These are required to be in the PKG-INFO file.
+REQUIRED_ATTRS = ('name', 'version', 'summary', 'home_page', 'author', 'author_email', 'license',
+                  'description', 'platform')
+
+
+class Package(FileContentUnit):
     """
     This class represents a Python package.
```
bmbouter: Do we want docstrings on the model definition? If we had docstrings here, it would allow the documentation pages about unit definitions be autogenerated. For this PR, this doesn't need a docstring before it is merged.

plugins/pulp_python/plugins/models.py
```diff
@@ -3,23 +3,58 @@
...
+    summary = StringField()
+
+    _checksum = StringField()
+    _checksum_type = StringField(default=DEFAULT_CHECKSUM_TYPE)
+    _filename = StringField()
+    _metadata_file = StringField()
+
+    # For backward compatibility
+    _ns = StringField(default='units_python_package')
+    _content_type_id = StringField(required=True, default=constants.PACKAGE_TYPE_ID)
```
bmbouter: FWIW, I've been definition the string value here and remove `PACKAGE_TYPE_ID` from `constants`. I then refer to it as `Package._content_type_id.default`

plugins/pulp_python/plugins/models.py
mhrivnak: During the sync, md5 is only used as the verification algorithm. Then the call to ``models.Package.from_archive``, made from the ``DownloadPackagesStep.download_succeeded`` method, re-calculates the checksum using the default algorithm (sha512) and saves that. I think I preserved previous behavior with all of this, but let me know if you think there's a change. If not, I'm inclined to leave it.

plugins/pulp_python/plugins/importers/importer.py
bmbouter: Oh that is right, good deal. FYI: https://pulp.plan.io/issues/1467

plugins/pulp_python/plugins/models.py
bmbouter: I see now, so the `sha512` is used elsewhere. Let's leave this as is then.

bmbouter:
Does there need to be any migrations for this? The data layer should stay the same, but what about dropping old indexes since new ones will be created that may be slightly different. Typically the other conversions have had some index migration work and docs made for them. I'm not sure in this case.

bmbouter:
Once merged, the Python models should be added to the [validation_script](https://github.com/pulp/pulp/blob/master/playpen/mongoengine/validation_check.py). Perhaps the validation script should be refactored to test with all models as discovered through entry points? That would handle this problem more generally across the plugins without having to update the validation script each time. What do you think?

bmbouter:
I'm done w/ the review. There are some comments to handle, but I don't need to re-review as long as you're satisfied. I'm not sure why travis and jenkins aren't happy, but my local tests ran fine so I think it's ok to merge red in this case.

plugins/pulp_python/plugins/models.py
mhrivnak: I don't feel strongly about it.

plugins/pulp_python/plugins/models.py
mhrivnak: I actually like the idea that a new checksum should only be calculated deliberately, hopefully right after the content has been verified some other way. In the python package case, we verify with an md5 checksum, then calculate a new sha512 checksum.

It would be a shame to have a situation where there is no checksum provided, the file is somehow corrupt, and we end up automatically saving a new checksum of the corrupt file without realizing it.

plugins/pulp_python/plugins/models.py
mhrivnak: In this case, the constant is in ``pulp_python.common`` and also gets used by the pulp-admin extensions, so we need to keep it. I think that's a reasonable approach.

plugins/pulp_python/plugins/importers/importer.py
mhrivnak: Good idea. Since @pcreech is working on doing this in the platform now, I'm going to let him handle it there.

mhrivnak:
There were only two indexes present before the conversion, and they are both created identically by mongoengine. There is an additional index created by mongoengine, which it does create successfully during upgrade. So I think we're ok without a migration.

mhrivnak:
I think this is ready to merge, but I'd like @rbarlow to at least take a glance at it first.

plugins/pulp_python/plugins/importers/importer.py
bmbouter: :+1:

plugins/pulp_python/plugins/models.py
bmbouter: I agree, thanks for explaining this. Given the explanation let's leave as-is.

plugins/pulp_python/plugins/models.py
bmbouter: I agree and will keep this in mind for other plugin types. Thanks for the explanation

mhrivnak:
ok test

plugins/pulp_python/plugins/importers/importer.py
rbarlow: Do you want to drop this line since it is being added to platform?

plugins/pulp_python/plugins/importers/sync.py
```diff
@@ -23,11 +25,18 @@
...
+    This DownloadStep subclass contains the code to process the downloaded manifests and determine
+    what is available from the feed. It does this as it gets each metadata file to spread the load
+    on the database.
     """
 
+    def __init__(self, *args, **kwargs):
+        """
+        Injects download requests into the call to the parent class's __init__
+        """
+        kwargs['downloads'] = self.generate_download_requests()
```
rbarlow: Since downloads is a required param, it should be listed explicitly in the method parameters and have the standard docblock entries.

plugins/pulp_python/plugins/models.py
rbarlow: I like the idea, but agree that it's not needed for this PR. @bmbouter want to file a general issue about this for all Pulp models? We can do that when you become the documentation czar!

plugins/pulp_python/plugins/models.py
rbarlow: I kind of regret that I didn't make another field for "feed_checksum" and keep the md5 there and use that for uniqueness during a sync, but use sha512 during publishing. You shouldn't do that in this PR obviously, but that could be a nice improvement (perhaps in all of our models?)

rbarlow: I also kind of regret making the checksum be two fields instead of just using a notation like algo:hash and storing that in one field. This is one thing I do like about how Docker did it!

plugins/pulp_python/plugins/models.py
bmbouter: Yes I will file this issue now -- "the documentation czar"

plugins/pulp_python/plugins/models.py
```diff
@@ -183,48 +180,9 @@ def _metadata_label(attribute):
...
-
     def __repr__(self):
         """
-        Return a string representation of self.
-
         :return: A string representing self.
         :rtype:  basestring
         """
-        return 'Python Package: %(name)s-%(version)s' % {'name': self.name, 'version': self.version}
+        return 'Package(name={}, version={})'.format(self.name, self.version)
```
rbarlow: It might be good to leave the "Python " on the front of this string, even though the class is called Package, since this is for human readability. It probably doesn't appear anywhere to users, but it could appear in debug logs?

plugins/test/unit/plugins/importers/test_sync.py
```diff
@@ -1013,20 +813,20 @@ def test_generate_download_requests(self):
         request_urls = [r.url for r in requests]
         self.assertEqual(
             request_urls,
-            ['http://example.com/cool.tar.gz', 'http://example.com/beats.tar.gz'])
+            ['http://u1/foo-1.2.0.tar.gz', 'http://u2/foo-1.3.0.tar.gz'])
```
rbarlow: I'm sad that you got rid of cool beats.

rbarlow:
I think you should make that downloads parameter be explicit before merging, but other than that LGTM.

plugins/pulp_python/plugins/models.py
bmbouter: I filed it as [Task 1522](https://pulp.plan.io/issues/1522).

plugins/pulp_python/plugins/importers/importer.py
mhrivnak: Good call. It made it into platform now, so I'll remove it.

plugins/pulp_python/plugins/importers/sync.py
mhrivnak: ``downloads`` is not a required param. In fact if a caller did pass it in, we overwrite that value here.

plugins/pulp_python/plugins/models.py
mhrivnak: Good thought. I do like sticking to the guidance on keeping ``repr`` close to something that could be passed to ``eval``, but something more user-friendly would be good in a ``__str__`` method.

plugins/pulp_python/plugins/importers/sync.py
rbarlow: On Thu, Jan 14, 2016 at 06:23:54AM -0800, Michael Hrivnak wrote:
> > +        kwargs['downloads'] = self.generate_download_requests()
> 
> ``downloads`` is not a required param. In fact if a caller did pass it in, we overwrite that value here.

Ugh, I really hope I only wrote that because I hadn't had enough coffee
yet. Speaking of which, is it… coffee time?

-- 
Randy Barlow
irc:   bowlofeggs

mhrivnak:
Scratch builds passed on rhel6, rhel7, and f23.

PR #60, opened 2016-01-24
Now that #293 is fixed, we don't need this empty config.

rbarlow:


rbarlow:
Neither of the test systems' failures are due to this pull request, as they also happen to other pull requests.

PR #62, opened 2016-02-01
Fix broken http config

pcreech:
pulp_puppet.conf was removed, but dev script was still creating
a link to it.  Removing the link.

closes #[1611](https://pulp.plan.io/issues/1611)
https://pulp.plan.io/issues/1611


Ichimonji10:
:+1:

PR #64, opened 2016-02-05
This uniqueness contstraint is now enforced by the platform for all c…

dkliban:
…ontent units.

https://pulp.plan.io/issues/1477
re: 1477


dkliban:
ok test

PR #66, opened 2016-02-24
Adds check for duplicate unit key

dkliban:
re #1406

### DIFF
--- a/plugins/pulp_python/plugins/importers/importer.py
+++ b/plugins/pulp_python/plugins/importers/importer.py
@@ -3,6 +3,8 @@ from itertools import chain
 import shutil
 import tempfile
 
+from mongoengine import NotUniqueError
+
 from pulp.plugins.importer import Importer
 from pulp.server.controllers import repository as repo_controller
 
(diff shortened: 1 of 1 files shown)


plugins/pulp_python/plugins/importers/importer.py
```diff
@@ -183,7 +185,11 @@ def upload_unit(self, repo, type_id, unit_key, metadata, file_path, conduit, con
         :rtype:           dict
         """
         package = models.Package.from_archive(file_path)
-        package.save_and_import_content(file_path)
+        try:
+            package.save_and_import_content(file_path)
+        except NotUniqueError:
+            package = package.__class__.objects.filter(**package.unit_key).first()
```
rbarlow: You can also use .get() instead of .filter().first() which will get you one object. However, get() does raise an Exception if nothing matches.

rbarlow:
I filed an issue about the Jenkins failure. It's not due to your change:

https://pulp.plan.io/issues/1719

Travis has been failing for a long time. It might be worth just turning off.

plugins/pulp_python/plugins/importers/importer.py
rbarlow: What about the uploaded file? Where will it go? Will it get garbage collected by the task working dir mechanism? I'll assume yes, just want to make sure.

rbarlow:
Before you merge, be sure to add a unit test for this. pulp_python used to have 100% coverage. Not sure it still does, but it might be good to strive for. Alternatively, add a pulp-smash test.

rbarlow:
To reiterate, on your other PRs I suggested filing an issue to test later was fine. However, as this project has historically had 100% test coverage I think we should strive to maintain it. That's my opinion though. If others disagree I can go with a vote on it.

rbarlow:
I just checked and the master branch currently does have 100% coverage, so we should definitely maintain it.

PR #68, opened 2016-02-25
Fix the travis build.

rbarlow:


rbarlow:
The Python 2.7 Travis build says that there is missing coverage on a blank line and a method signature (?) and the Python 2.6 Travis build seems to be missing dependencies for Celery or something, but still reports 100% coverage (?). I spent a lot of time on this today and got to the point that I don't think it's worth further investment at this time.

PR #69, opened 2016-03-01
Configure Travis to only flake8 and pep257 the code.

rbarlow:
We have spent too much time maintaining our travis file. It
regularly breaks and is often difficult to debug. Since we
primarily use Jenkins, we don't really need Travis though it does
provide some additional value. The current problems are taking too
long to sort out to justify the time, so this commit adjusts it to
only run flake8 and pep257 on the code. We can re-enable it later
if we have the time and desire to do so.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,25 +2,8 @@ language: python
 python:
   - "2.6"
   - "2.7"
-before_install:
-  - "pushd .."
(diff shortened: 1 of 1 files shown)


seandst:
You sure you don't wanna remove pep257? :)

PR #74, opened 2016-04-27
Migrations need to be run when upgrading to 1.1.1

seandst:


docs/release-notes/1.1.rst
```diff
@@ -1,6 +1,13 @@
 1.1 Release Notes
 =================
 
+1.1.1
+-----
+This release includes migrations, so the Upgrade procedure documented below
+in the 1.1.0 release notes should be run.
```
asmacdo: A little awkward. How about:

Migrations will need to be run as part of the upgrade to this release. This procedure is documented below in the 1.1.0 release notes.

seandst: Will steal this verbatim, thanks.

PR #75, opened 2016-05-10
Create new model for python packages

asmacdo:
Please note, this is being merged to a feature branch which will contain this commit and others to get sync and publish working. Because this is only changing the model, it is expected that any tests outside of `plugins/test/unit/plugins/test_models.py` may fail.

The filename contains the name, version, platform, and python versions.
This means that filename is guaranteed to be unique on PyPI.
Additionally, this commit removes some unecessary field names to prevent
them from being wrongly populated with data from a later release, which
will occur if the packages are built from the JSON metadata from PyPI.

closes #1882

### DIFF
--- a/plugins/pulp_python/plugins/migrations/0002_smaller_model_for_json.py
+++ b/plugins/pulp_python/plugins/migrations/0002_smaller_model_for_json.py
@@ -1,0 +1,22 @@
+"""
+This migration removes `home_page`, `platform`, 'author_email`, `description` and `license` from
+the units_python_package collection. Additionally, it renames `_filename` to `filename`.
+"""
+from pulp.server.db import connection
+
+
+def migrate(*args, **kwargs):
+    """
+    Perform the migration as described in this module's docblock.
+    """
+    db = connection.get_database()
+    collection = db['units_python_package']
+
+    collection.update({}, {"$unset": {"home_page": True}}, multi=True)
+    collection.update({}, {"$unset": {"platform": True}}, multi=True)
+    collection.update({}, {"$unset": {"license": True}}, multi=True)
+    collection.update({}, {"$unset": {"_metadata_file": True}}, multi=True)
+    collection.update({}, {"$unset": {"description": True}}, multi=True)
+    collection.update({}, {"$unset": {"author_email": True}}, multi=True)
+
+    collection.update({}, {"$rename": {"_filename": "filename"}}, multi=True)
--- a/plugins/pulp_python/plugins/models.py
+++ b/plugins/pulp_python/plugins/models.py
@@ -1,120 +1,124 @@
-from gettext import gettext as _
 import hashlib
-import re
-import tarfile
 
 from mongoengine import StringField
 from pulp.server.db.model import FileContentUnit
+import twine  # noqa
 
 from pulp_python.common import constants
 
 
 DEFAULT_CHECKSUM_TYPE = 'sha512'
-
-
-# These are required to be in the PKG-INFO file.
-REQUIRED_ATTRS = ('name', 'version', 'summary', 'home_page', 'author', 'author_email', 'license',
-                  'description', 'platform')
+REQUIRED_ATTRS = ('name', 'version', 'author', 'summary')
 
 
 class Package(FileContentUnit):
     """
     This class represents a Python package.
+
+    Packages stored on PyPI have significantly more metadata associated with them, but we have
+    chosen to keep this model minimal. Below is an explanation for why this was done.
+
+    There are some overloaded terms that should be briefly discussed. We will be using the
+    semantics of PEP 426. https://www.python.org/dev/peps/pep-0426/#id14
+
+    Package - An individual, installable python package. A package is defined by the project,
+              target architecture, and Python version. Note that there may not be a target
+              architecture in the case of architecture agnostic packages. A Python version is also
+              optional if there are no requirements for a specific runtime version. There are 3
+              packaging formats: sdist, whl, and zip.
+    Release - A snapshot of the project. A particular release of the project will all share the
+              same version number, but there may be multiple packages for a given release for
+              different architectures, python versions, and package types.
+    Project - The entire group of packages of all releases.
+
+    The way that PyPI stores the metadata on their end leaves this model open to potential
+    inaccuracies. Many packages of the same project share a significant amount of metadata,
+    including mutable fields. In the situation where a mutable field changes between releases, only
+    the latest version is present in the metadata accessible over the API. This is could be
+    particularly problematic for a field like "license". For this reason, most of these fields have
+    been left out of this model. The only fields that are included here are necessary for sync
+    (name, version, filename) or are useful for disambiguation between similarly named projects
+    (author, summary). By keeping the scope of the metadata as small as is reasonable, we have less
+    complexity and reduced risk of innacuracy.
+
+    :ivar name: name of the project, ex. scipy
+    :type name: basestring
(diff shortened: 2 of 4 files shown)


plugins/pulp_python/plugins/migrations/0002_smaller_model_for_json.py
```diff
@@ -0,0 +1,20 @@
+"""
+This migration removes `home_page`, `platform`, and `license` from the units_python_package
+collection. Additionally, it renames `_filename` to `filename` and makes it the pk.
```
jeremycline: It eventually clicked that pk was primary key, but I'd prefer it to be spelled out for me.

plugins/pulp_python/plugins/models.py
```diff
@@ -1,120 +1,85 @@
...
+    author = StringField(required=True)
     author_email = StringField()
     description = StringField()
-    home_page = StringField()
-    license = StringField()
-    platform = StringField()
     summary = StringField()
 
     _checksum = StringField()
     _checksum_type = StringField(default=DEFAULT_CHECKSUM_TYPE)
```
jeremycline: I guess it's outside the scope of this PR, but I really wish this was just "checksum" and in the format "<type>:<hex_value>". It's done this way on all the other units, though, so I suppose we should be consistent.

plugins/pulp_python/plugins/models.py
```diff
@@ -1,120 +1,85 @@
...
-
-# These are required to be in the PKG-INFO file.
-REQUIRED_ATTRS = ('name', 'version', 'summary', 'home_page', 'author', 'author_email', 'license',
-                  'description', 'platform')
+REQUIRED_ATTRS = ('name', 'version', 'author', 'author_email', 'description', 'summary')
 
 
 class Package(FileContentUnit):
     """
     This class represents a Python package.
```
jeremycline: It would be great if you documented this class more extensively. I know why you are using the filename as the unit key, but only because you told me in person. I'd like this doc-block to explain what each field is, where it comes from, and why it is there. You've done a lot of research I don't want to have to do myself :smile:

plugins/pulp_python/plugins/models.py
```diff
@@ -1,120 +1,85 @@
-from gettext import gettext as _
 import hashlib
-import re
-import tarfile
 
 from mongoengine import StringField
 from pulp.server.db.model import FileContentUnit
+import twine  # noqa
```
asmacdo: @jeremycline am I crazy? I had to add noqa here because flake8 said this was the incorrect order...

jeremycline: That is weird. I feel like someone else had a similar issue with import ordering recently, but I can't recall who.

plugins/pulp_python/plugins/models.py
```diff
@@ -1,120 +1,85 @@
...
-        :return:             An instance of Package that represents the package found at
-                             archive_path.
-        :rtype:              pulp_python.plugins.models.Package
-        :raises:             ValueError if archive_path does not point to a valid Python tarball
-                             created with setup.py sdist.
-        :raises:             IOError if the archive_path does not exist.
+        Create and return (but do not save) an instance of a single python package object from the
+        metadata available from PyPI JSON.
+
+        :param file_data: metadata specific to a file
```
jeremycline: I'm not exactly sure what this means. Which file? At first I thought it was from the JSON from the PyPI API, but I see there's a second dictionary, distribution_data. How do they differ and where do they come from?

plugins/pulp_python/plugins/models.py
```diff
@@ -1,120 +1,85 @@
...
+        :param release: version number
+        :type  release: basestring
+        :param distribution_data: metadata that applies to all versions
+        :type  distribution_data: dict
+        :return: instance of a package
+        :rtype:  pulp_python.plugins.models.Package
+        """
+        package_attrs = {}
+        package_attrs['filename'] = file_data['filename']
+        package_attrs['name'] = distribution_data['name']
```
jeremycline: Couldn't you just drop ``name`` into the list you're iterating over below?

plugins/pulp_python/plugins/models.py
```diff
@@ -1,120 +1,85 @@
...
+        from the package itself.
+
+        :param path: path to the package
+        :type  path: basestring
+        :return: instance of a package
+        :rtype:  pulp_python.plugins.models.Package
+        """
+
+        meta_dict = twine.package.PackageFile.from_filename(path, comment='').metadata_dictionary()
+        filtered_dict = {}
```
jeremycline: I dream of the day we can use dictionary comprehensions.

jeremycline:
Just documentation comments, really. If you want me to proof-read your docs I am happy to do that, otherwise merge when you're ready.

asmacdo:
@jeremycline, I added more docs, removed a couple more fields. I would appreciate if you would let me know if the new documentation makes sense and sounds rational. I kept the commit separate to make it easy for you to see. I'll rebase before merge.

plugins/pulp_python/plugins/models.py
```diff
@@ -8,20 +8,54 @@
...
 DEFAULT_CHECKSUM_TYPE = 'sha512'
-REQUIRED_ATTRS = ('name', 'version', 'author', 'author_email', 'description', 'summary')
+REQUIRED_ATTRS = ('name', 'version', 'author', 'summary')
 
 
 class Package(FileContentUnit):
     """
     This class represents a Python package.
+
+    Packages stored on PyPI have significantly more metadata associated with them but we have
```
jeremycline: Comma before the "but"

plugins/pulp_python/plugins/models.py
```diff
@@ -8,20 +8,54 @@
...
+    chosen to keep this model minimal. Below is an explanation for why this was done.
+
+    There are some overloaded terms that should be briefly discussed. We will be using the
+    semantics of PEP 426. https://www.python.org/dev/peps/pep-0426/#id14
+
+    Package - An individual, installable python package. A package is of a specific version, and
+              may also require a specific architecture and python version. Additionally, there
+              may be seperate packages for each package type (sdist, whl, zip).
```
jeremycline: Maybe:
```
Package - An individual, installable Python package. A package is defined by the version of
    the project, target architecture, and Python version. Note that there may not be a target
    architecture in the case of architecture-agnostic packages. A Python version is also
    optional if there are no requirements for a specific runtime version. There are three
    packaging formats: sdist, whl, and zip.
```
I'm not convinced the above is better, so pick and choose as you like. If you stick with what you have, ``s/seperate/separate``

plugins/pulp_python/plugins/models.py
```diff
@@ -8,20 +8,54 @@
...
+    a field like "license". For this reason, most of these fields have been left out of this model.
+    The only fields that are included here are necessary for sync (name, version, filename) or are
+    useful for disambiguation between similarly named packages (author, summary). By keeping the
+    scope of the metadata as small as is reasonable, we have a less complexity and reduced risk of
+    innacuracy.
+
+    :ivar name: name of the project, ex. scipy
+    :type name: basestring
+    :ivar version: Contains the distribution's version number. This field must be in the format
+                   specified in PEP 386 .
```
jeremycline: There's a space before the period.

plugins/pulp_python/plugins/models.py
```diff
@@ -8,20 +8,54 @@
...
+    :type version: basestring
+    :ivar filename: name of the file containaing the package and metadata
```
jeremycline: s/containaing/containing

PR #77, opened 2016-05-20
Change sync code to use new model

asmacdo:
Adds some fields that were necessary for sync, syncs in the new way, and
heavily modify docblocks to explain.

https://pulp.plan.io/issues/135

### DIFF
--- a/plugins/pulp_python/plugins/models.py
+++ b/plugins/pulp_python/plugins/models.py
@@ -41,24 +41,33 @@ class Package(FileContentUnit):
     (author, summary). By keeping the scope of the metadata as small as is reasonable, we have less
     complexity and reduced risk of innacuracy.
 
+    :ivar author: primary author of the package
+    :type author: basestring
+    :ivar filename: name of the file containing the package and metadata
+    :type filename: basestring
+    :ivar md5_digest: md5 checksum provided by PyPI to ensure we got the correct bits
+    :type md5_digest: basestring
     :ivar name: name of the project, ex. scipy
     :type name: basestring
+    :ivar packagetype: format of python package, ex bdist_wheel, sdist
+    :type packagetype: basestring
+    :ivar url: url that the package can be downloaded from
+    :type url: basestring
     :ivar version: Contains the distribution's version number. This field must be in the format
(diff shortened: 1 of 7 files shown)


plugins/pulp_python/plugins/models.py
```diff
@@ -41,24 +41,33 @@ class Package(FileContentUnit):
...
     :ivar summary: one line summary of what the package does
     :type summary: basestring
     """
 
-    name = StringField(required=True)
-    version = StringField(required=True)
-    filename = StringField(required=True)
     author = StringField()
+    filename = StringField(required=True)
+    md5_digest = StringField()
```
jeremycline: I'm not crazy about having the md5_digest and _checksum and _checksum_type.

While you're in here, can we change this to a dictionary called ``checksums`` or something where the keys are the checksum algorithm and the value is the hex digest? I think this is also happening in the pulp_rpm plugin so if you do this make sure they match. It's not as nice as unifying our file and checksum modeling, but it'll be consistent at least.

jeremycline:
Just the one comment. Since this is going into a feature branch and will be reviewed as a whole, I didn't do a complete examination to make sure you didn't miss a ``_filename`` to ``filename``, and I'm okay with the code having TODOs in it.

plugins/pulp_python/plugins/models.py
asmacdo: The reason I added this field is because _checksum and _checksum_type are both "Pulp fields", but md5_digest is a pure "Python field" in that it is a mirror of the field from PyPI. We generate our own checksum for later (not md5). I thought that it seemed like it made sense to save the PyPI checksum as well as our own, so that the client would be in a position to check that it matches both. Additionally, looking forward, when we are doing Pulp-Pulp syncs we can use either checksum, but I at least wanted the option of being able to naively syncing from another Pulp server without having to do anything different than we would I we were syncing from PyPI.

That all said, I think that it would be possible (and acceptable) to remove the md5_digest and save only a single checksum onto the object. I would like to have the pulp-pulp sync working before spending too much effort on that decision. Lets make sure to have a discussion before the final merge of the feature branch.

PR #79, opened 2016-06-01
Refactor to process uploaded files with twine

asmacdo:
By generating the unit key (which for Python is the filename), we are
able to rename the tempororary file (the filename is a random uuid)
to the actual filename. This allows twine to determine the file type
so it can be processed accordingly.

closes #136


asmacdo:
https://pulp.plan.io/issues/136

PR #80, opened 2016-06-03
Publish metadata to allow pulp to pulp sync

asmacdo:
Publishes a json metadata file for each python project that contains
metadata for each python package that belongs to it. This is a subset of
the api that PyPI implements, and allows a Pulp to sync from another
Pulp in exactly the same way that it would sync from PyPI itself.

closes #140

### DIFF
--- a/plugins/pulp_python/plugins/distributors/steps.py
+++ b/plugins/pulp_python/plugins/distributors/steps.py
@@ -1,9 +1,12 @@
+import collections
 from gettext import gettext as _
 import itertools
+import json
 import logging
 import os
 from xml.etree import cElementTree as ElementTree
 
+import pkg_resources
 from pulp.plugins.util.publish_step import AtomicDirectoryPublishStep, PluginStep
 from pulp.server.controllers import repository as repo_controller
 
@@ -33,7 +36,7 @@ class PublishContentStep(PluginStep):
         """
         Publish all the python files themselves by creating the symlinks to the storage paths.
         """
-        for name, packages in _get_packages(self.get_conduit().repo_id).items():
+        for name, packages in _get_projects(self.get_conduit().repo_id).items():
             for package in packages:
                 relative_path = _get_package_path(name, package['filename'])
                 symlink_path = os.path.join(self.parent.web_working_dir, relative_path)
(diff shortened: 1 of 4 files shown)


asmacdo:
https://pulp.plan.io/issues/140

plugins/pulp_python/plugins/distributors/steps.py
```diff
@@ -76,14 +80,22 @@ def process_main(self):
...
                 element.text = name
                 ElementTree.SubElement(body, 'br')
                 PublishMetadataStep._create_package_index(name, simple_path, packages)
 
             index.write(ElementTree.tostring(html, 'utf8'))
 
+        for name, packages in projects.items():
+            project_metadata_path = os.path.join(api_path, name, 'json')
+            os.makedirs(project_metadata_path)
+            project_index_metadata_path = os.path.join(project_metadata_path, 'index.html')
```
jeremycline: Oof. It's confusing to have a file that's called index.html that just contains JSON.

plugins/pulp_python/plugins/distributors/steps.py
```diff
@@ -76,14 +80,22 @@ def process_main(self):
...
+            with open(project_index_metadata_path, 'w') as meta_json:
```
jeremycline: Will this actually cover every case? Reading https://wiki.python.org/moin/PyPIJSON leads me to believe I should be able to get the JSON from ``project_metadata_path``, but with this setup it looks like it's only available at ``project_metadata_path/index.html``. Both work with PyPI.

plugins/pulp_python/plugins/distributors/steps.py
```diff
@@ -121,6 +133,55 @@ def _create_package_index(name, simple_path, packages):
 
             package_index.write(ElementTree.tostring(html, 'utf8'))
 
+    @staticmethod
+    def _create_metadata(name, packages):
```
jeremycline: I feel like there should be a model for ``Project``, and the ``Package`` model should have a [many-to-one](https://docs.djangoproject.com/en/1.9/topics/db/models/#many-to-one-relationships) relationship with a ``Project``. This could be an instance method of ``Project``. @seandst, does that sound crazy?

plugins/pulp_python/plugins/distributors/steps.py
asmacdo: https://github.com/pulp/pulp/pull/2607 should allow this to work.

plugins/pulp_python/plugins/distributors/steps.py
seandst: I'm coming to this without any context. So project is a pypi project, which can have multiple packages, but packages can only have one Project? If so, the relationship is indeed one project to many packages, which in pulp 3 would be a foreign key relationship from package back to project.

PR #84, opened 2016-06-28
Sync publish and upload all types that are compatible with PyPI

asmacdo:


extensions_admin/pulp_python/extensions/admin/upload.py
```diff
@@ -21,16 +21,21 @@ def determine_type_id(self, filename, **kwargs):
...
+        :type  filename: basestring
+        :param args:     Unused
+        :type  args:     list
+        :param kwargs:   Unused
+        :type  kwargs:   dict
+        :return:         An empty dictionary
+        :rtype:          dict
         """
-        return {}
+        return {"filename": filename.split('/')[-1]}
```
jeremycline: I _think_ this will work in all cases (on unixy filesystems, anyway), but ``os.path.basename`` does this as well and is immediately obvious.

plugins/pulp_python/plugins/distributors/steps.py
```diff
@@ -33,13 +33,12 @@ def process_main(self):
...
-            for package in packages:
-                relative_path = _get_package_path(name, package['filename'])
-                symlink_path = os.path.join(self.parent.web_working_dir, relative_path)
-                if not os.path.exists(os.path.dirname(symlink_path)):
-                    os.makedirs(os.path.dirname(symlink_path))
-                os.symlink(package['storage_path'], symlink_path)
+        repo_id = self.get_conduit().repo_id
+        for package in models.Package.objects.packages_in_repo(repo_id):
+            symlink_path = os.path.join(self.parent.web_working_dir, 'packages', package.src_path)
+            if not os.path.exists(os.path.dirname(symlink_path)):
```
jeremycline: It's unfortunate the exception in Python 2 is OSError since that's annoying to handle compared to FileExistsError in Python 3. Otherwise I'd recommend ``try/except`` this rather than wrapping it in an ``if``. Ah well.

plugins/pulp_python/plugins/models.py
```diff
@@ -137,43 +167,60 @@ def checksum(path, algorithm=DEFAULT_CHECKSUM_TYPE):
...
+        """
+        return pkg_resources.parse_version(self.version)
 
-    @staticmethod
-    def _metadata_label(attribute):
+    @property
+    def src_path(self):
         """
-        Return the label in the PKG-INFO file that corresponds to the given attribute.
+        Returns the relative path to the package bits.
```
jeremycline: Might not hurt to have an example here. I eventually got that it'd be something like ``source/p/pkginfo/<whatever-the-filename-format-is>``, but I had to jump around a bit.

plugins/pulp_python/plugins/models.py
```diff
@@ -1,120 +1,150 @@
...
+    been left out of this model. The only fields that are included here are necessary for sync
+    (name, version, filename, path) or are useful for disambiguation between similarly named
+    projects (author, summary). By keeping the scope of the metadata as small as is reasonable, we
+    have less complexity and reduced risk of innacuracy.
+
+    :ivar author: primary author of the package
+    :type author: basestring
+    :ivar filename: name of the file containing the package and metadata
+    :type filename: basestring
+    :ivar md5_digest: md5 checksum provided by PyPI to ensure we got the correct bits
```
jeremycline: I haven't looked at how this is used yet, but it makes me nervous since PEP 426 only states the source URL SHOULD include a hash and it RECOMMENDS one that is guaranteed by the current  standard library ``hashlib``[0]. It just so happens that PyPI uses md5 sums right now, but that could change. I might have stronger feelings (one way or the other) when I see how the field is used.

[0] https://www.python.org/dev/peps/pep-0426/#source-url

plugins/pulp_python/plugins/models.py
```diff
@@ -1,120 +1,150 @@
...
-    home_page = StringField()
-    license = StringField()
-    platform = StringField()
+    filename = StringField(required=True)
+    md5_digest = StringField()
+    name = StringField(required=True)
+    packagetype = StringField()
+    path = StringField()
     summary = StringField()
+    version = StringField(required=True)
```
jeremycline: Do you think it's worth performing validation on the version string based on https://www.python.org/dev/peps/pep-0440/#version-scheme?

plugins/pulp_python/plugins/models.py
```diff
@@ -1,120 +1,150 @@
...
+    :type version: basestring
+    """
 
     author = StringField()
-    author_email = StringField()
-    description = StringField()
-    home_page = StringField()
-    license = StringField()
-    platform = StringField()
+    filename = StringField(required=True)
```
jeremycline: I haven't seen one, but is there a formal definition of the accept filename format in a PEP somewhere? It'd be nice to have a validator for this as well.

plugins/pulp_python/plugins/models.py
```diff
@@ -1,120 +1,150 @@
...
+    md5_digest = StringField()
+    name = StringField(required=True)
```
jeremycline: It may be worth slapping a validator on this field using https://www.python.org/dev/peps/pep-0426/#name (sans the bit about "confusable" characters since I don't think people will be using Pulp to host a public index server, but maybe?)

plugins/pulp_python/plugins/models.py
```diff
@@ -1,120 +1,150 @@
...
+    packagetype = StringField()
```
jeremycline: It may be worth defining the set supported package types and validating this field, too.

plugins/pulp_python/plugins/models.py
```diff
@@ -1,120 +1,150 @@
...
-            attrs['_filename'] = '%s-%s.tar%s' % (attrs['name'], attrs['version'], compression_type)
-            attrs['_checksum'] = checksum
-            attrs['_checksum_type'] = DEFAULT_CHECKSUM_TYPE
-            attrs['_metadata_file'] = metadata_file_name
-            package = cls(**attrs)
-            return package
-        finally:
-            if 'package_archive' in locals():
-                package_archive.close()
+    def from_json(cls, package_data, release, project_data):
```
jeremycline: If/when PEP 426 is accepted, my interpretation is that the JSON served by a package index implementing the 2.0 version of the metadata should match the defined schema[0]. It might be worth leaving a note on this method that one day it might want to validate against that schema.

[0] https://www.python.org/dev/peps/pep-0426/#metadata-validation

plugins/pulp_python/plugins/models.py
jeremycline: I see now that the json PyPI currently serves up has md5_digest as a field. I guess in the future we'd need to pull the hash off the ``source_url`` field and it could be anything.

plugins/pulp_python/plugins/models.py
```diff
@@ -1,120 +1,150 @@
...
+        package_attrs['name'] = project_data['name']
+        package_attrs['author'] = project_data['author']
+        package_attrs['summary'] = project_data['summary']
+
+        package_attrs['filename'] = package_data['filename']
+        package_attrs['path'] = package_data['path']
+        package_attrs['packagetype'] = package_data['packagetype']
+        package_attrs['md5_digest'] = package_data['md5_digest']
+
+        # If we are syncing from PyPI, there will be no `checksum`, but will be `md5_digest`
```
jeremycline: Based on https://github.com/pypa/warehouse/pull/891 I _think_ we can't expect an ``md5_digest`` when warehouse is released, but we can expect one or more ``{digest_type}_digest`` fields (currently just md5 and sha256 it looks like).

plugins/pulp_python/plugins/models.py
```diff
@@ -1,120 +1,150 @@
...
+    """
 
     author = StringField()
-    author_email = StringField()
-    description = StringField()
-    home_page = StringField()
-    license = StringField()
-    platform = StringField()
+    filename = StringField(required=True)
+    md5_digest = StringField()
```
jeremycline: While we're validating everything, it couldn't hurt to slap on a validator here. A regex accepting 64 characters that are [a-f][0-9] and ignores case would be fine (I think).

plugins/pulp_python/plugins/models.py
```diff
@@ -1,120 +1,150 @@
...
+
+        Twine is smart enough to crack open a tarball, zip, or wheel and parse the metadata that is
+        contained in the package.
+
+        :param path: path to the package
+        :type  path: basestring
+        :return: instance of a package
+        :rtype:  pulp_python.plugins.models.Package
+        """
+        meta_dict = PackageFile.from_filename(path, comment='').metadata_dictionary()
```
jeremycline: The metadata dictionary coming back from twine looks to include at least ``md5_digest`` (in the future I expect there will be ``sha256_digest`` as well). It would be good if, at this point, the file was confirmed to match the checksum in the metadata. On https://github.com/pulp/pulp_python/pull/84/files#diff-3905b2a2875b436d8a5d0e7818c6826eR146 the checksum is calculated again and it could be that this is performed on a corrupt file, right?

https://github.com/pypa/twine/blob/a37fcbd2e4631144f1de2912d4ee99bc3d5d1bb8/twine/package.py#L92

plugins/pulp_python/plugins/models.py
```diff
@@ -1,120 +1,150 @@
...
+        filtered_dict = {}
+        # Use only a subset of the attributes in the metadata to keep package minimal.
+        for key, value in meta_dict.iteritems():
+            if key in PACKAGE_ATTRS:
+                filtered_dict[key] = value
+        filtered_dict['filename'] = path.split('/')[-1]
```
jeremycline: You could use ``os.path.basename`` here

plugins/pulp_python/plugins/models.py
```diff
@@ -1,120 +1,150 @@
...
+        :param path: path to the package
+        :type  path: basestring
+        :return: instance of a package
+        :rtype:  pulp_python.plugins.models.Package
+        """
+        meta_dict = PackageFile.from_filename(path, comment='').metadata_dictionary()
+        filtered_dict = {}
+        # Use only a subset of the attributes in the metadata to keep package minimal.
+        for key, value in meta_dict.iteritems():
+            if key in PACKAGE_ATTRS:
```
jeremycline: I see ``md5_digest`` isn't in ``PACKAGE_ATTRS``. I actually think that's probably for the best since it may be a different digest type in the future, but I wonder if we want to make sure at least one checksum is provided by the metadata and use it.

plugins/pulp_python/plugins/models.py
```diff
@@ -137,43 +167,60 @@ def checksum(path, algorithm=DEFAULT_CHECKSUM_TYPE):
...
+    @property
+    def package_specific_metadata(self):
+        """
+        Returns a dictionary containing the subset of metadata that is not duplicated between
+        packages of the same project.
+
+        :return: metadata for package that is not shared with the rest of the project
+        :rtype:  dict
+        """
+        href = '../../../packages/%s' % self.checksum_url
```
jeremycline: This looks very "magic-y" to me. The fact that it backs up three directories is a bit of a mystery to me when I read this. My assumption is that this metadata is only valid when published to a certain place. Could you document why it's  three directories and where this metadata is valid?

plugins/pulp_python/plugins/models.py
```diff
@@ -1,120 +1,150 @@
...
+
+    The way that PyPI stores the metadata on their end leaves this model open to potential
+    inaccuracies because they assume that all packages of the same project share certain metadata,
+    including mutable fields. In the situation where a mutable field changes between releases, only
+    the latest version is present in the metadata accessible over the API. This is could be
+    particularly problematic for a field like "license". For this reason, most of these fields have
+    been left out of this model. The only fields that are included here are necessary for sync
+    (name, version, filename, path) or are useful for disambiguation between similarly named
+    projects (author, summary). By keeping the scope of the metadata as small as is reasonable, we
+    have less complexity and reduced risk of innacuracy.
```
jeremycline: I made a note on ``package_specific_metadata`` about documenting the section of metadata it provides. It might be worth documenting what the expected PyPI JSON is for this model.

plugins/pulp_python/plugins/querysets.py
```diff
@@ -0,0 +1,39 @@
...
+
+        :param repo_id: Identifies the repository from which to retrieve packages
+        :type  repo_id: basestring
+        :return: Dictionary containing packages in the repo, organized by project name
+        :rtype:  dict
+        """
+        packages = self.packages_in_repo(repo_id)
+        projects = {}
+
+        for pkg in packages:
```
jeremycline: This could potentially be a very, very big dictionary in memory, right? I'll have to look at how this gets used, but it would be nice if it could use a generator.

plugins/pulp_python/plugins/importers/sync.py
```diff
@@ -39,117 +39,103 @@ def __init__(self, *args, **kwargs):
...
-        for u in manifest_urls:
+        metadata_urls = [urljoin(self.parent._feed_url, 'pypi/%s/json/' % pn)
+                         for pn in self.parent._project_names]
+        for u in metadata_urls:
             if self.canceled:
                 return
             yield request.DownloadRequest(u, StringIO(), {})
 
-    def _process_manifest(self, manifest):
+    def _process_metadata(self, metadata):
```
jeremycline: Do you think it would be reasonable to put this as a class method on the package and have return a list of Package objects? It could be called ``from_metadata`` or similar.

plugins/pulp_python/plugins/importers/sync.py
```diff
@@ -39,117 +39,103 @@ def __init__(self, *args, **kwargs):
...
+        if checksum != package._checksum:
             report.state = 'failed'
-            report.error_report = {'expected_checksum': report.data._checksum,
+            report.error_report = {'expected_checksum': package._checksum,
                                    'actual_checksum': checksum}
             return self.download_failed(report)
 
-        package = models.Package.from_archive(report.destination)
+        # Unless checksum from upstream is default type, recalculate checksum
+        if package._checksum_type is not models.DEFAULT_CHECKSUM_TYPE:
```
jeremycline: I think you want ``!=`` here

plugins/pulp_python/plugins/importers/sync.py
```diff
@@ -39,117 +39,103 @@ def __init__(self, *args, **kwargs):
...
+        checksum = models.Package.checksum(report.destination, package._checksum_type)
+        if checksum != package._checksum:
             report.state = 'failed'
-            report.error_report = {'expected_checksum': report.data._checksum,
+            report.error_report = {'expected_checksum': package._checksum,
                                    'actual_checksum': checksum}
             return self.download_failed(report)
 
-        package = models.Package.from_archive(report.destination)
+        # Unless checksum from upstream is default type, recalculate checksum
```
jeremycline: I'm a little surprised by this behavior. Doesn't this mean ``DEFAULT_CHECKSUM_TYPE`` is more like ``CHECKSUM_TYPE``? I expect a default to be used if nothing else is available, but not if there is something else set.

plugins/pulp_python/plugins/importers/sync.py
```diff
@@ -206,16 +193,15 @@ def __init__(self, repo, conduit, config, working_dir):
...
-        DownloadRequest for its url attribute.
+        Yield a Nectar Download request for each package that wasn't available locally.
 
         :return: A generator that yields DownloadReqests for the Package files.
         :rtype:  generator
         """
-        for p in self.get_local_units_step.units_to_download:
-            url = self.unit_urls.pop(p)
+        for package in self.get_local_units_step.units_to_download:
+            url = os.path.join(self._feed_url, 'packages', package.path)
```
jeremycline: It looks like the metadata includes a url field. Even if we don't keep that on the model, we could have a property that creates this url.

plugins/pulp_python/plugins/distributors/steps.py
```diff
@@ -60,67 +59,125 @@ def process_main(self):
...
-            title.text = 'Links for %s' % name
+            title.text = 'Links for %s' % project_name
             ElementTree.SubElement(head, 'meta', {'name': 'api-version', 'value': '2'})
             body = ElementTree.SubElement(html, 'body')
             heading = ElementTree.SubElement(body, 'h1')
             heading.text = title.text
             for package in packages:
-                href = '../../%s#%s=%s' % (_get_package_path(name, package['filename']),
-                                           package['checksum_type'], package['checksum'])
+                href = '../../packages/%s' % package.checksum_url
```
jeremycline: If you opt to add a method to the model to build this url, you can replace it here I think.

plugins/pulp_python/plugins/distributors/steps.py
```diff
@@ -60,67 +59,125 @@ def process_main(self):
...
+        :type  projects: dict
+        """
+        api_path = os.path.join(self.parent.web_working_dir, 'pypi')
+        for project_name, packages in projects.items():
+            project_metadata_path = os.path.join(api_path, project_name, 'json')
+            os.makedirs(project_metadata_path)
+            project_index_metadata_path = os.path.join(project_metadata_path, 'index.json')
+            with open(project_index_metadata_path, 'w') as project_json:
+                data = PublishMetadataStep._create_project_metadata(project_name, packages)
+                project_json.write(json.dumps(data))
```
jeremycline: This works, but ``json`` provides the ``dump`` method which does the same thing. It's also nice if people will ever look at this metadata to format it prettily. This is purely my preference and you can take it or leave it :shipit:.

``json.dump(data, project_json, sort_keys=True, indent=4, separators=(',', ': '))``

plugins/pulp_python/plugins/importers/sync.py
asmacdo: It seems weird to me to have a method on the model that returns more than one instance of that model.

plugins/pulp_python/plugins/querysets.py
asmacdo: I think you are correct. Would it be  better to query for all unique package names in the repo and then iterate through those, creating a new query for all packages of that name in the repo? I was trying to avoid so many db calls, but that is probably better than destroying the ram.

plugins/pulp_python/plugins/importers/sync.py
jeremycline: It is sort of weird. I thought about maybe suggesting putting it on the QuerySet (reasoning is that it's sort of like a Django Manager and those typically handle operations involving many instances of a model). Even though that _also_ seems a little weird, maybe it's the right place for it.

https://docs.djangoproject.com/en/1.9/topics/db/managers/#adding-extra-manager-methods

plugins/pulp_python/plugins/querysets.py
jeremycline: It's this sort of thing that makes me feel like maybe we should be explicitly modeling the concept of ``Project`` and use it to relate packages. It just feels cleaner to me. I want to ruminate on it a little more :cow:.

plugins/pulp_python/plugins/distributors/steps.py
jeremycline: Actually you might want to handle it with a try/except and check the OSError's errno, since there is a race condition here between two processes publishing repos with a share directory structure.

plugins/pulp_python/plugins/models.py
asmacdo: Digging into `pkg_resources.parse_version` shows that if the version is valid, it is used. If it is invalid, it is a legacy version and is still supported. In warehouse, new data will be validated, legacy data will be grandfathered in.

My stance is that if it works on PyPI, we should support it, even if it does not follow the PEP. For that reason, I am going to leave validation of version out, and will also leave a comment in the code explaining why.

plugins/pulp_python/plugins/models.py
asmacdo: Looks like there is legacy data for filename and project name as well. I think we should just let it all in.

plugins/pulp_python/plugins/models.py
asmacdo: I don't think that it 426 will be accepted: https://www.python.org/dev/peps/pep-0426/#pep-deferral

Anyway, that refers to a `metadata.json` file that lives in the package. It might be used by twine, but I think they use `METADATA` instead.
I have seen metadata versions on PyPI spanning from 1.0 to 2.0, but this is data stored in the package itself (like

plugins/pulp_python/plugins/querysets.py
asmacdo: I might be convinced as well.

jeremycline: I did sleep on this and unrelated to whether or not we have a ``Project`` model, this could probably be written to return a generator of tuples. Something like ``(name, (list, of, packages))``. However, I'm happy spend the weekend thinking about what we want to do here.

plugins/pulp_python/plugins/importers/sync.py
asmacdo: I agree with what you have said, and have played with my options to see if this will work. The biggest problem is that we cannot import the model into the QuerySet due to circular imports, so I would rather leave it as it is.

jeremycline: Ah. That is unfortunate. How do you feel about a ``Project`` model, then? It sounded like you were considering it, and this could live on that model for now.

plugins/pulp_python/plugins/models.py
```diff
@@ -1,123 +1,155 @@
...
 
+    # Validation deliberately left out because of noncompliant legacy data that works on PyPI
     author = StringField()
-    author_email = StringField()
-    description = StringField()
-    home_page = StringField()
-    license = StringField()
-    platform = StringField()
+    filename = StringField(required=True)
+    md5_digest = StringField(regex=r'[a-f0-9]{32}')
```
asmacdo: A-Z

plugins/pulp_python/plugins/models.py
```diff
@@ -222,6 +227,18 @@ def project_metadata(self):
         """
         return {'name': self.name, 'summary': self.summary, 'author': self.author}
 
+    def feed_url(self, feed_url):
```
jeremycline: maybe ``package_url`` would be a nicer name?

asmacdo: agreed.

jeremycline:
I trust you'll make Jenkies happy before you merge, but you have my blessing

ipanova:
ok test

ipanova:
ok test

PR #86, opened 2016-07-26
Remove flake8-import-order

asmacdo:
Fixes broken test runner for systems that install flake8 3.0 (including fedora 24 vagrant dev environment).

flake8 3.0 introduced a backwards incompatible change in the config that
breaks some flake8 plugins including flake8-import-order. Since this
plugin has also had some occasional unexpected behavior, we should just
remove it for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,5 +5,5 @@ python:
 install:
   - "pip install -r test_requirements.txt"
 script:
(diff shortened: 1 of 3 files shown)


asmacdo:
This failed travis because flake8 3.0 dropped support for Python 2.6.

https://pulp.plan.io/issues/2106
https://github.com/pulp/pulp_python/pull/87

PR #88, opened 2016-07-26
pin to flake8-2.6.2 for py2.6 support

asmacdo:
https://pulp.plan.io/issues/2106


asmacdo:
ok test

asmacdo:
flake8-import-order seems to have dropped python 2.6 compatibility between 0.8 and 0.9. Rather than pin that also, I am just removing it.

PR #96, opened 2016-10-13
Fix the path field in json publish

asmacdo:
A refactor near the end of the review process removed the need for `../../../packages`, on the sync side, but remained in the code path to publish.

### DIFF
--- a/plugins/pulp_python/plugins/models.py
+++ b/plugins/pulp_python/plugins/models.py
@@ -211,10 +211,9 @@ class Package(FileContentUnit):
         :return: metadata for package that is not shared with the rest of the project
         :rtype:  dict
         """
-        href = '../../../packages/%s' % self.checksum_path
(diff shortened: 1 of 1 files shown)


plugins/pulp_python/plugins/models.py
```diff
@@ -211,10 +211,9 @@ def package_specific_metadata(self):
         :return: metadata for package that is not shared with the rest of the project
         :rtype:  dict
         """
-        href = '../../../packages/%s' % self.checksum_path
-        return {'filename': self.filename, 'packagetype': self.packagetype, 'path': href,
-                'md5_digest': self.md5_digest, 'checksum': self._checksum,
-                'checksum_type': self._checksum_type}
+        return {'filename': self.filename, 'packagetype': self.packagetype,
```
ipanova: @asmacdo seems like you need to update the tests, fix them and good to merge

PR #98, opened 2017-01-30
Problem: release notes for 2.0.0 are nonexistent

dkliban:
Solution: Add release notes for 2.0.0

closes #2537
https://pulp.plan.io/issues/2537

### DIFF
--- a/docs/release-notes/2.0.rst
+++ b/docs/release-notes/2.0.rst
@@ -1,0 +1,27 @@
+2.0 Release Notes
+=================
+
+2.0.0
+-----
+
+This release adds ability to manage Python wheel packages. Wheel is currently considered the
+standard for built and binary packaging for Python. More information about the differences between
+sdist and wheel formats can be found in the `Python Packaging User Guide
+<https://packaging.python.org/wheel_egg/>`_. Wheel packages can now be both uploaded and synced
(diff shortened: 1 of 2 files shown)


docs/release-notes/2.0.rst
```diff
@@ -0,0 +1,49 @@
...
+<https://packaging.python.org/wheel_egg/>`_. Wheel packages can now be both uploaded and synced
+from PyPi. When syncing from PyPi both sdist and wheel packages are synced. This results in many
+packages existing in both formats in the Pulp Python repositories. This behavior maintains support
+for older pip clients that are not aware of the wheel format.
+
+Requirements
+^^^^^^^^^^^^
+The 2.0.x releases of the Python plugins depend on the Pulp 2.12.x or greater.
+
+Upgrade
```
seandst: Do we have any generic upgrade docs to link to here? If so, we should link to them. If not, we've recently started adding a separate note about restarting goferd and pulp-streamer only if they're installed.

seandst: I also wonder about just leaving the upgrade section out, but I'm not sure that's appropriate...

dkliban: I don't mind leaving out the upgrade section all together.

PR #99, opened 2017-02-18
Modify directive name to generate docs correctly using platfrom config

goosemania:
re #2585
https://pulp.plan.io/issues/2585


goosemania:
Wrong branch

PR #104, opened 2017-04-04
Add provides for python2-* packages

pcreech:
Add provides for python2-* packages.  Fedora downstream packages are
named python2-* and provide python-*.  Upstream needs to mirror this for
dependency resolution to work properly

re #2687


dralley:
ok test

PR #110, opened 2017-07-13
add MSI and RPM as valid package types

mikeadamz:
Fixes https://pulp.plan.io/issues/2899


asmacdo:
The types added here are not going to be fully supported (no uploads) but PyPI still manages these legacy types, so we should too. Thanks for digging in @mikeadamz, great work!

PR #111, opened 2017-07-13
Update AUTHORS

asmacdo:
Also alphabetize


mikeadamz:
Mike Adams <mike.adams@unix-security.net>

PR #112, opened 2017-07-13
Update documentation to include additional requirements for syncing Python repos between Pulp Instances

mikeadamz:


docs/user-docs/getting_started.rst
```diff
@@ -185,3 +185,7 @@ used as a feed for a second Pulp. Here is an example.::
    $ pulp-admin python repo create --repo-id sync_from_other_pulp --feed http://<host>/pulp/python/web/<repo_id>/ --package-names numpy,scipy
 
    $ pulp-admin python repo sync run --repo-id sync_from_other_pulp
+
+The pulp_python RPM distribution automatically adds index.json as a DirectoryIndex in pulp_python.conf. If you're not running the pulp_python RPM distribution, you will need to ensure this setting is present before remote Pulp instances will be able to sync your repository.
```
asmacdo: I think this info would fit better https://github.com/mikeadamz/pulp_python/blob/0ac47451a77d5a65eb3827599ddc354c4d9ce490/docs/admin-docs/installation.rst#server

bmbouter:
@mikeadamz Also when squashing can the commit message refer to issue 2900 and have better line also in the commit msg. We have docs on how to do that here:  http://docs.pulpproject.org/dev-guide/contributing/branching.html#commit-messages

Thank you!

docs/user-docs/getting_started.rst
mikeadamz: Maybe add it to both places? 99% of the time the getting started doc is as far as I go. Had this documentation existed in the server config doc I probably wouldn't have even read it.

mikeadamz:
I deleted my original pulp_python repository like a dummy because I thought it would be quicker than rolling back my repo to squash the two commits into one, but didn't realize i wouldn't be able to update this pull request to point it back at the new repository.. Opening PR #113 to fix

PR #115, opened 2017-09-15
Sync succeeds even when feed or package-names is not specified

ammaritiz:
An appropriate error message will now be displayed
when a user tries to sync a python repo that is
missing feed or package-names

closes #2964


werwty:
ok test

PR #116, opened 2017-10-31
Bootstrap pulp_python plugin

dralley:
Everything needed for getting the plugin to be installed properly by Pulp 3, and get basic CRUD functionality working.


COPYRIGHT
```diff
@@ -1,4 +1,4 @@
-Copyright © 2014 Pulp Project developers.
+Copyright © 2017 Pulp Project developers.
```
dralley: Would it be best to leave this at 2014?  I don't think it matters one way or the other in this case

werwty: I think this should be left as 2014: 

> The general rule is that the year to include in a copyright notice is the year of first publication of the work. [0]

[0] http://www.copyrightlaws.com/copyright-notice-year/

docs/Makefile
```diff
@@ -1,153 +0,0 @@
-# Makefile for Sphinx documentation
```
werwty: I think eventually we're still planning on building sphinx docs. Although I am fine with this being deleted for now and bought back when we have sphinx docs to build.

docs/release-notes/1.0.rst
```diff
@@ -1,51 +0,0 @@
-1.0 Release Notes
```
werwty: Same for these. I think we will need to bring the release notes back at some point, but it's fine if they go for now.

pulp_python/app/models.py
```diff
@@ -0,0 +1,84 @@
+"""
```
werwty: Can this doc string be removed?

pulp_python/app/serializers.py
```diff
@@ -0,0 +1,75 @@
+"""
+Check `Plugin Writer's Guide`_ and `pulp_example`_ plugin
```
werwty: Can this doc string be removed?

pulp_python/app/viewsets.py
```diff
@@ -0,0 +1,53 @@
+"""
+Check `Plugin Writer's Guide`_ and `pulp_example`_ plugin
```
werwty: Can this doc string also be removed?

flake8.cfg
```diff
@@ -1,5 +1,5 @@
 [flake8]
-exclude = ./docs/*
+exclude = */migrations/*
```
werwty: I think we should keep excluding docs

PR #118, opened 2017-12-07
Create models for PythonPackageContent

werwty:
closes #2883
https://pulp.plan.io/issues/2883

### DIFF
--- a/pulp_python/app/models.py
+++ b/pulp_python/app/models.py
@@ -1,35 +1,65 @@
-from gettext import gettext as _
 from logging import getLogger
 
 from django.db import models
 
-from pulpcore.plugin.models import (Artifact, Content, ContentArtifact, RemoteArtifact, Importer,
-                                    ProgressBar, Publisher, RepositoryContent, PublishedArtifact,
-                                    PublishedMetadata)
-from pulpcore.plugin.tasking import Task
-
+from pulpcore.plugin.models import Content, Importer, Publisher
+from django.contrib.postgres.fields import ArrayField
 
 log = getLogger(__name__)
 
 
-class PythonContent(Content):
+class Classifier(models.Model):
     """
-    The "python" content type.
+    Custom tags for classifier
 
-    Define fields you need for your new content type and
-    specify uniqueness constraint to identify unit of this type.
+    Fields:
 
-    For example::
+        name (models.TextField): The name of the classifier
 
-        field1 = models.TextField()
-        field2 = models.IntegerField()
-        field3 = models.CharField()
+    Relations:
 
-        class Meta:
-            unique_together = (field1, field2)
+        python_package_content (models.ForeignKey): The PythonPackageContent this classifier
+        is associated with.
+
+    """
+
+    name = models.TextField()
+    python_package_content = models.ForeignKey("PythonPackageContent", related_name="classifiers",
+                                               related_query_name="classifier")
+
+
+class PythonPackageContent(Content):
+    """
+    A Content Type representing Python's Distribution Package as
+    defined in pep-0426 and pep-0345
+    https://www.python.org/dev/peps/pep-0491/
+    https://www.python.org/dev/peps/pep-0345/
     """
 
     TYPE = 'python'
+    filename = models.TextField(unique=True, db_index=True, blank=False)
+    packagetype = models.TextField(blank=False)
+    name = models.TextField(blank=False)
+    version = models.TextField(blank=False)
+    metadata_version = models.TextField(blank=False)
+    summary = models.TextField()
+    description = models.TextField()
+    keywords = ArrayField(models.TextField())
+    home_page = models.TextField()
+    download_url = models.TextField()
+    author = models.TextField()
+    author_email = models.TextField()
+    maintainer = models.TextField()
+    maintainer_email = models.TextField()
+    license = models.TextField()
+    requires_python = models.TextField()
+    project_url = models.TextField()
+    platform = models.TextField()
+    supported_platform = models.TextField()
+    requires_dist = ArrayField(models.TextField())
+    provides_dist = ArrayField(models.TextField())
+    obsoletes_dist = ArrayField(models.TextField())
+    requires_external = ArrayField(models.TextField())
 
 
 class PythonPublisher(Publisher):
--- a/pulp_python/app/serializers.py
+++ b/pulp_python/app/serializers.py
@@ -1,30 +1,164 @@
+from gettext import gettext as _
+
 from rest_framework import serializers
 from pulpcore.plugin import serializers as platform
 
 from . import models
 
 
-class PythonContentSerializer(platform.ContentSerializer):
+class ClassifierSerializer(serializers.ModelSerializer):
     """
-    A Serializer for PythonContent.
-
-    Add serializers for the new fields defined in PythonContent and
-    add those fields to the Meta class keeping fields from the parent class as well.
-
-    For example::
-
-    field1 = serializers.TextField()
-    field2 = serializers.IntegerField()
-    field3 = serializers.CharField()
-
-    class Meta:
-        fields = platform.ContentSerializer.Meta.fields + ('field1', 'field2', 'field3')
-        model = models.PythonContent
+    A serializer for Python Classifiers
     """
 
+    name = serializers.CharField(
+        help_text=_("A string giving a single classification value for a Python package.")
+    )
+
     class Meta:
-        fields = platform.ContentSerializer.Meta.fields
-        model = models.PythonContent
+        model = models.Classifier
+        fields = ('name',)
+
+
+class PythonPackageContentSerializer(platform.ContentSerializer):
+    """
+    A Serializer for PythonPackageContent.
+    """
+
+    filename = serializers.CharField(
+        help_text=_('The name of the distribution package, usually of the format:'
+                    ' {distribution}-{version}(-{build tag})?-{python tag}-{abi tag}'
+                    '-{platform tag}.{packagetype}')
+    )
+    packagetype = serializers.CharField(
+        help_text=_('The type of the distribution package '
+                    '(e.g. sdist, bdist_wheel, bdist_egg, etc)')
+    )
+    name = serializers.CharField(
+        help_text=_('The name of the python project.')
+    )
+    version = serializers.CharField(
+        help_text=_('The packages version number.')
+    )
+    metadata_version = serializers.CharField(
+        help_text=_('Version of the file format')
+    )
+    summary = serializers.CharField(
+        required=False, allow_blank=True,
+        help_text=_('A one-line summary of what the package does.')
+    )
+    description = serializers.CharField(
+        required=False, allow_blank=True,
+        help_text=_('A longer description of the package that can run to several paragraphs.')
+    )
+    keywords = serializers.ListField(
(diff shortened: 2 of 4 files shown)


werwty:
## Testing
### Creating a Content Unit
1. Create an Artifact
`http --form POST http://0.0.0.0:8000/api/v3/artifacts/ file@./scipy.tar.gz`
2. Create a Content Unit
scipy.json
```
{
  "project_url": "https://www.scipy.org",
  "description": "SciPy (pronounced Sigh Pie) is open-source software for mathematics, science, and engineering. The SciPy library depends on NumPy, which provides convenient and fast N-dimensional array manipulation. The SciPy library is built to work with NumPy arrays, and provides many user-friendly and efficient numerical routines such as routines for numerical integration and optimization. Together, they run on all popular operating systems, are quick to install, and are free of charge. NumPy and SciPy are easy to use, but powerful enough to be depended upon by some of the world's leading scientists and engineers. If you need to manipulate numbers on a computer and display or publish the results, give SciPy a try!",
  "home_page": "https://www.scipy.org",
  "maintainer_email": "",
  "requires_external": "",
  "path": "19/b1/a3ea10ee5425ca3c04f63aba1bb72e4d8f5535db99389016e980063238ac/scipy-0.9.0-cp26-cp26mu-manylinux1_x86_64.whl",
  "artifacts": {"scipy.tar.gz":"http://0.0.0.0:8000/api/v3/artifacts/47026179-2c33-4960-ad12-c4e8a8cea089/"},
  "provides_dist": "",
  "packagetype": "sdist",
  "obsoletes_dist": "",
  "requires_dist": "numpy (>=1.8.2)",
  "version": "1.0.0",
  "classifiers": [
  {"name": "Development Status :: 5 - Production/Stable"},{"name": "Intended Audience :: Developers"}
    ],
  "supported_platform": "",
  "metadata_version": "2.0",
  "maintainer": "",
  "requires_python": ">=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*",
  "license": "BSD",
  "platform": "Windows",
  "download_url": "https://github.com/scipy/scipy/releases",
  "name": "scipy",
  "keywords": "string",
  "author_email": "scipy-dev@python.org",
  "summary": "SciPy: Scientific Library for Python",
  "filename": "scipy-0.9.0-cp26-cp26mu-manylinux1_x86_64.whl",
  "author": "SciPy Developers"
}
```
`http POST http://0.0.0.0:8000/api/v3/content/python/ < scipy.json`

3.) Test Filtering
`http GET http://0.0.0.0:8000/api/v3/content/python/?name=&version=&filename=&classifier=Intended+Audience+%3A%3A+Developers`

pulp_python/app/models.py
```diff
@@ -1,35 +1,73 @@
...
     """
 
     TYPE = 'python'
+    filename = models.TextField(unique=True, db_index=True, blank=False)
+    path = models.TextField(blank=False)
+    packagetype = models.TextField(blank=False)
+    name = models.TextField(blank=False)
+    version = models.TextField(blank=False)
+
+    # Note that this field is not contained in PyPI nor Warehouse
```
asmacdo: Could you clarify this comment? This field isn't returned by the json API, but it is a part of the PKG-INFO/METADATA .

pulp_python/app/serializers.py
```diff
@@ -1,30 +1,87 @@
...
+    A Serializer for PythonPackageContent.
     """
 
+    filename = serializers.CharField()
+    path = serializers.CharField()
+    packagetype = serializers.CharField()
+    name = serializers.CharField()
+    version = serializers.CharField()
+    metadata_version = serializers.CharField()
+    summary = serializers.CharField(required=False, allow_blank=True)
```
asmacdo: I don't want to pile on a time consuming task, but these all need help text. If you prefer, we can have decide them together on a call and do that as a separate task.

pulp_python/app/models.py
```diff
@@ -1,35 +1,73 @@
...
+    Relations:
 
-        class Meta:
-            unique_together = (field1, field2)
+        pythonpackagecontent (models.ForeignKey): The task this tag is associated with
+
+    """
+
+    name = models.TextField()
+    pythonpackagecontent = models.ForeignKey("PythonPackageContent", related_name="classifiers",
```
asmacdo: I think `python_package_content`.

pulp_python/app/models.py
```diff
@@ -1,35 +1,73 @@
...
+    license = models.TextField()
+    requires_python = models.TextField()
+    project_url = models.URLField()
+    platform = models.TextField()
+
+    # Note supported_platform exists in PEP0345 but not in legacy-PyPI or warehouse
+    # ignore this on publish to PyPI or Warehouse
+    supported_platform = models.TextField()
+
+    # These should be reference to other packages, treat them as text fields for now
```
asmacdo: I don't think these can reference other packages really. For instance, scipy returns "numpy (>=1.8.2)" for "requires_dist". I think text is all we can do now, and probably all we can do for now.

pulp_python/app/models.py
```diff
@@ -1,35 +1,73 @@
...
+
+    name = models.TextField()
+    pythonpackagecontent = models.ForeignKey("PythonPackageContent", related_name="classifiers",
+                                             related_query_name="classifier")
+
+
+class PythonPackageContent(Content):
+    """
+    A Content Type representing Python's Distribution Package as
+    defined in pep-0426
```
asmacdo: Link to PEP 426 would be nice.

asmacdo: Also 345?

pulp_python/app/models.py
werwty: neither warehouse nor PyPI stores this information (that I could tell)

pulp_python/app/models.py
werwty: I think we should be able to do dependency solving. It's my hope that we can specify "sync this and all dependencies" to pulp. An have this be links to the various dependencies.

pulp_python/app/viewsets.py
```diff
@@ -1,22 +1,31 @@
...
 
 
-class PythonContentViewSet(platform.ContentViewSet):
+class PythonPackageContentFilter(filterset.FilterSet):
     """
-    A ViewSet for PythonContent.
+    Filter for PythonPackageContent
+    """
+
+    classifier = BaseInFilter(name='classifier__name', lookup_expr='in')
```
asmacdo: I don't think "in" is what users would want. ?classifier=Topic :: Scientific/Engineering,Topic :: Software Development (except properly escaped) would return an *or* of those tags. 

However we do it, I think we should add filtering in a separate story.

pulp_python/app/serializers.py
```diff
@@ -1,30 +1,87 @@
...
+        PythonPackageContent = models.PythonPackageContent.objects.create(**validated_data)
+        for classifier in classifiers:
+            models.Classifier.objects.create(pythonpackagecontent=PythonPackageContent,
+                                             **classifier)
+        return PythonPackageContent
+
     class Meta:
-        fields = platform.ContentSerializer.Meta.fields
-        model = models.PythonContent
+        fields = platform.ContentSerializer.Meta.fields + ('filename', 'path', 'packagetype',
```
asmacdo: Just personally preference, but for a list this long, I think start it on the next line.

    fields = platform.ContentSerializer.Meta.fields + (
        'filename', 'path', 'packagetype',  ... 
        ...
     )

pulp_python/app/models.py
asmacdo: The trouble is that ">" can't be a link. These are closer to queries than links.

pulp_python/app/models.py
```diff
@@ -23,51 +23,42 @@ class Classifier(models.Model):
...
     A Content Type representing Python's Distribution Package as
-    defined in pep-0426
+    defined in pep-0426 and pep-0345
+    https://www.python.org/dev/peps/pep-0491/
+    https://www.python.org/dev/peps/pep-0345/
     """
 
     TYPE = 'python'
     filename = models.TextField(unique=True, db_index=True, blank=False)
-    path = models.TextField(blank=False)
```
werwty: Removed path, this is something not part of the metadata. The information is the same as Pulp's artifact relative_path

pulp_python/app/models.py
```diff
@@ -23,51 +23,42 @@ class Classifier(models.Model):
...
-
     summary = models.TextField()
     description = models.TextField()
-    keywords = models.TextField()
+    keywords = ArrayField(models.TextField())
     home_page = models.URLField()
     download_url = models.URLField()
     author = models.TextField()
-    author_email = models.EmailField()
+    author_email = models.TextField()
```
werwty: It turns out that a valid email only needs to conform to [RFC822](http://www.faqs.org/rfcs/rfc822.html)
so `Author-email: "C. Schultz" <cschultz@example.com>` is valid for python metadata, but django fails the email validation for it.

pulp_python/app/models.py
```diff
@@ -23,51 +23,42 @@ class Classifier(models.Model):
...
     maintainer = models.TextField()
-    maintainer_email = models.EmailField()
+    maintainer_email = models.TextField()
     license = models.TextField()
     requires_python = models.TextField()
-    project_url = models.URLField()
+    project_url = models.TextField()
```
werwty: python metadata is expecting: 
> A string containing a browsable URL for the project and a label for it, separated by a comma.
Which fails django url validation

pulp_python/app/models.py
```diff
@@ -1,35 +1,64 @@
...
-    specify uniqueness constraint to identify unit of this type.
+        name (models.TextField): The name of the classifier
 
-    For example::
+    Relations:
 
-        field1 = models.TextField()
-        field2 = models.IntegerField()
-        field3 = models.CharField()
+        pythonpackagecontent (models.ForeignKey): The task this tag is associated with
```
asmacdo: python_package_content here too

pulp_python/app/models.py
```diff
@@ -1,35 +1,64 @@
...
     TYPE = 'python'
+    filename = models.TextField(unique=True, db_index=True, blank=False)
+    packagetype = models.TextField(blank=False)
+    name = models.TextField(blank=False)
+    version = models.TextField(blank=False)
+    metadata_version = models.TextField(blank=False)
+    summary = models.TextField()
+    description = models.TextField()
+    keywords = ArrayField(models.TextField())
+    home_page = models.URLField()
```
asmacdo: Do these other URL fields have the same problem? Rather than worry about it, maybe just make them all text fields, and we can fiddle with this when we do validations.

pulp_python/app/serializers.py
```diff
@@ -1,30 +1,164 @@
...
+        help_text=_('A list of additional keywords to be used to assist searching for the '
+                    'package in a larger catalog.')
+    )
+    home_page = serializers.URLField(
+        required=False, allow_blank=True,
+        help_text=_('The URL for the package\'s home page.')
+    )
+    download_url = serializers.URLField(
+        required=False, allow_blank=True,
+        help_text=_('The URL from which this version of the package can be downloaded. ')
```
asmacdo: + from Pulp. In the past, there has been confusion about the `path` (where you can download this from the remote sourcel) and `download_url` (where you can download this from Pulp)

PR #119, opened 2017-12-11
Add sync to pulp_python

werwty:
## Testing:

Create Importer:
`http POST http://localhost:8000/api/v3/importers/python/ name='bar' download_policy='immediate' sync_mode='mirror' feed_url='https://test.pypi.org' repository='http://localhost:8000/api/v3/repositories/ed57db7c-16f2-416e-b4df-b12b372fe9a5/' projects:='["pulpcore", "Django"]'`

Sync:
`http POST http://localhost:8000/api/v3/importers/python/{importer-id}/sync/`

## TODO

- [x] unittests
- [x] doc strings
- [x] flake8 failures
- [ ] figure out sane max_length default

### DIFF
--- a/pulp_python/app/models.py
+++ b/pulp_python/app/models.py
@@ -1,11 +1,25 @@
-from logging import getLogger
+import json
+import os
 
+from collections import namedtuple
+from logging import getLogger
+from urllib.parse import urljoin, urlparse
+
+from django.contrib.postgres.fields import ArrayField
 from django.db import models
 
-from pulpcore.plugin.models import Content, Importer, Publisher
-from django.contrib.postgres.fields import ArrayField
+from pulpcore.plugin.models import (Artifact, Content, Importer, Publisher)
+
+from pulpcore.plugin.changeset import (
+    ChangeSet,
+    PendingArtifact,
+    PendingContent,
+    SizedIterable
+)
 
 log = getLogger(__name__)
+
+Delta = namedtuple('Delta', ('additions', 'removals'))
 
 
 class Classifier(models.Model):
@@ -41,25 +55,25 @@ class PythonPackageContent(Content):
     packagetype = models.TextField(blank=False)
     name = models.TextField(blank=False)
     version = models.TextField(blank=False)
-    metadata_version = models.TextField(blank=False)
-    summary = models.TextField()
-    description = models.TextField()
-    keywords = ArrayField(models.TextField())
-    home_page = models.TextField()
-    download_url = models.TextField()
-    author = models.TextField()
-    author_email = models.TextField()
-    maintainer = models.TextField()
-    maintainer_email = models.TextField()
-    license = models.TextField()
-    requires_python = models.TextField()
-    project_url = models.TextField()
-    platform = models.TextField()
-    supported_platform = models.TextField()
-    requires_dist = ArrayField(models.TextField())
-    provides_dist = ArrayField(models.TextField())
-    obsoletes_dist = ArrayField(models.TextField())
-    requires_external = ArrayField(models.TextField())
+    metadata_version = models.TextField(null=True)
+    summary = models.TextField(null=True)
+    description = models.TextField(null=True)
+    keywords = models.TextField(null=True)
+    home_page = models.TextField(null=True)
+    download_url = models.TextField(null=True)
+    author = models.TextField(null=True)
+    author_email = models.TextField(null=True)
+    maintainer = models.TextField(null=True)
+    maintainer_email = models.TextField(null=True)
+    license = models.TextField(null=True)
+    requires_python = models.TextField(null=True)
+    project_url = models.TextField(null=True)
+    platform = models.TextField(null=True)
+    supported_platform = models.TextField(null=True)
+    requires_dist = ArrayField(models.TextField(), default=[])
+    provides_dist = ArrayField(models.TextField(), default=[])
+    obsoletes_dist = ArrayField(models.TextField(), default=[])
+    requires_external = ArrayField(models.TextField(), default=[])
 
 
 class PythonPublisher(Publisher):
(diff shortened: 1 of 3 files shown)


pulp_python/app/models.py
```diff
@@ -42,20 +61,20 @@ class PythonPackageContent(Content):
...
-    author = models.TextField()
-    author_email = models.TextField()
-    maintainer = models.TextField()
-    maintainer_email = models.TextField()
-    license = models.TextField()
-    requires_python = models.TextField()
-    project_url = models.TextField()
-    platform = models.TextField()
-    supported_platform = models.TextField()
+    summary = models.TextField(null=True, max_length=8192)
```
asmacdo: Can you explain the reasoning for this length as a comment, even if it is just an arbitrary large number?

pulp_python/app/models.py
```diff
@@ -42,20 +61,20 @@ class PythonPackageContent(Content):
...
+    description = models.TextField(null=True, max_length=8192)
+    keywords = models.TextField(null=True)
```
asmacdo: For all these optional text fields, I think we want `blank=True`, not `null=True`.

https://docs.djangoproject.com/en/1.11/ref/models/fields/#null

pulp_python/app/models.py
```diff
@@ -92,9 +111,128 @@ class PythonImporter(Importer):
...
+
+        metadata_urls = [urljoin(self.feed_url, 'pypi/%s/json' % project)
+                         for project in self.project_names]
+
+
+        for metadata_url in metadata_urls:
+            parsed_url = urlparse(metadata_url)
+
+            download = self.get_futures_downloader(metadata_url, os.path.basename(parsed_url.path))
+            #download_result = download.fetch()
```
asmacdo: ?

pulp_python/app/models.py
```diff
@@ -92,9 +111,128 @@ class PythonImporter(Importer):
...
+        q_set = PythonPackageContent.objects.filter(repositories=self.repository)
+        q_set = q_set.only("filename")
+        for content in (c.cast() for c in q_set):
+            inventory.add(content.filename)
+        return inventory
+
+    def _fetch_remote(self):
+        remote = []
+
+        metadata_urls = [urljoin(self.feed_url, 'pypi/%s/json' % project)
```
asmacdo: This should be `pypi/%s/json/`. As it is will work for syncing from PyPI, but we need the trailing slash to sync from PyPI and other Pulps.

pulp_python/app/models.py
```diff
@@ -92,9 +111,128 @@ class PythonImporter(Importer):
...
+    def _parse_metadata(cls, project, version, distribution):
+
+        package = {}
+
+        package['filename'] = distribution['filename']
+        package['packagetype'] = distribution['packagetype']
+        package['name'] = project['name']
+        package['version'] = version
+        package['metadata_version'] = '3.0'
+        package['summary'] = project.get('summary')
```
asmacdo: For optional string fields (iff you make them blank=True), they should probably be,

    field = whatever.get('field') or ''

pulp_python/app/models.py
```diff
@@ -92,9 +111,128 @@ class PythonImporter(Importer):
...
+        package['requires_python'] = project.get('requires_python')
+        package['project_url'] = project.get('project_url')
+        package['platform'] = project.get('platform')
+        package['supported_platform'] = project.get('supported_platform')
+        package['requires_dist'] = project.get('requires_dist', [])
+        package['provides_dist'] = project.get('provides_dist', [])
+        package['obsoletes_dist'] = project.get('obsoletes_dist', [])
+        package['requires_external'] = project.get('requires_external', [])
+        package['url'] = distribution['url']
+        package['digest'] = distribution['digests'].get('sha256')
```
asmacdo: AFAIK, each distribution only has a `md5_digest` unless we are syncing from another Pulp

pulp_python/app/models.py
```diff
@@ -92,9 +111,128 @@ class PythonImporter(Importer):
...
+
+    def _parse_metadata(cls, project, version, distribution):
+
+        package = {}
+
+        package['filename'] = distribution['filename']
+        package['packagetype'] = distribution['packagetype']
+        package['name'] = project['name']
+        package['version'] = version
+        package['metadata_version'] = '3.0'
```
asmacdo: Not every distribution is 3.0. Since this field isn't included in the json, we should leave it blank and only populate it if we are getting the data straight out of a metadata file.

pulp_python/app/models.py
```diff
@@ -1,12 +1,31 @@
...
+from pulpcore.plugin.changeset import (
+    BatchIterator,
+    ChangeSet,
+    PendingArtifact,
+    PendingContent,
+    SizedIterable,
+)
+
 
 log = getLogger(__name__)
```
asmacdo: (nit) 2 newlines, not 1

pulp_python/app/models.py
```diff
@@ -1,12 +1,31 @@
...
 
+Delta = namedtuple('Delta', ('additions', 'removals'))
```
asmacdo: We ought to move this to core. This named tuple is part of the Plugin API, weird that each plugin has to write one.

pulp_python/app/models.py
```diff
@@ -92,9 +111,128 @@ class PythonImporter(Importer):
...
+            download()
+
+            metadata = json.load(open(download.writer.path))
+            for version, packages in metadata['releases'].items():
+                for package in packages:
+                    remote.append(self._parse_metadata(metadata['info'], version, package))
+
+        return remote
+
+    def _find_delta(self, inventory, remote, mirror=True):
```
asmacdo: I think we should default to "not destructive", so mirror=False.

pulp_python/app/models.py
werwty: I ran into an exception `DataError('value too long for type character varying(64)\n',)`
So to get sync working I just added a rather large max_length.

pulp_python/app/models.py
werwty: The issue is that some projects on pypi has the values set to null. 
E.g. `maintainer` in test pypi Django: https://test.pypi.org/pypi/Django/json

pulp_python/app/models.py
werwty: err the trailing slash does not work when accessing warehouse:
https://pypi.org/pypi/pulpcore/json vs https://pypi.org/pypi/pulpcore/json/

pulp_python/app/models.py
werwty: ah, so this is a difference between warehouse and legacy: Warehouse includes both the md5 and the sha256 by default

werwty: you know what, I think you are right, `md5_digest` is also present separately from the `digest` field from warehouse. If we use `md5_digest` it'll be backwards compatible to legacy. I'm convinced, will update.

werwty:
I'm going to hold off on figuring out a sane max_length default for now. I can figure them out when I do validations. But what this means for now is that there might be some failures associated with syncing projects with too long a description

pulp_python/app/models.py
```diff
@@ -41,25 +55,25 @@ class PythonPackageContent(Content):
...
-    license = models.TextField()
-    requires_python = models.TextField()
-    project_url = models.TextField()
-    platform = models.TextField()
-    supported_platform = models.TextField()
-    requires_dist = ArrayField(models.TextField())
-    provides_dist = ArrayField(models.TextField())
-    obsoletes_dist = ArrayField(models.TextField())
-    requires_external = ArrayField(models.TextField())
+    metadata_version = models.TextField(null=True)
```
werwty: I think we should keep the null values if they are set that way in the PyPI metadata, and not override it with a default ""

pulp_python/app/models.py
```diff
@@ -83,18 +97,171 @@ def publish(self):
...
+            q |= models.Q(pythonpackagecontent__filename=content_key)
+            q_set = self.repository.content.filter(q)
+            q_set = q_set.only('id')
+            for content in q_set:
+                yield content
 
     def sync(self):
         """
-        Synchronize the repository with the remote repository.
+        Sync the python projects listed on a :class:`pulp_python.app.models.Serializer`
```
asmacdo: I don't think you need to mention the serializer, the project list stored is on this object.

pulp_python/app/serializers.py
```diff
@@ -164,19 +164,16 @@ class Meta:
...
-    Similar to the example above, in PythonContentSerializer.
-    Additional validators can be added to the parent validators list
-
-    For example::
-
-    class Meta:
-        validators = platform.ImporterSerializer.Meta.validators + [myValidator1, myValidator2]
     """
 
+    project_names = serializers.ListField(
```
asmacdo: what do you think of calling this `requirements`? For now, just a list of project names, but in the future it could take specific versions or the other syntax in requirements.txt

pulp_python/app/serializers.py
```diff
@@ -164,19 +164,16 @@ class Meta:
...
+        child=serializers.CharField(),
+        required=False, default=[],
```
asmacdo: I think required=True. This importer only syncs right?

pulp_python/app/models.py
werwty: oops I totally meant importer there

PR #121, opened 2017-12-19
Update models to use TextField in lieu of ArrayField

werwty:
fixes #3217
https://pulp.plan.io/issues/3217


asmacdo:
could you also add fixes #1132? 
https://pulp.plan.io/issues/1132

PR #122, opened 2018-01-02
Added some simple validations and versioned repo support

werwty:


pulp_python/app/models.py
```diff
@@ -105,19 +109,23 @@ class PythonImporter(Importer):
     TYPE = 'python'
     projects = models.TextField(null=True)
 
-    def _fetch_inventory(self):
+    def _fetch_inventory(self, old_version):
         """
         Fetch existing contentunits in the repository.
 
+        Args:
+           old_version (pulpcore.plugin.models.RepositoryVersion): the latest pre-existing version
```
dralley: I don't think this should be called ```old_version``` if it is the latest existing version.  Maybe ```current_version``` (as opposed to "new_version" or something).

pulp_python/app/models.py
```diff
@@ -229,28 +240,39 @@ def _build_additions(self, additions, remote_metadata):
                 })
             yield content
 
-    def _build_removals(self, removals):
+    def _build_removals(self, removals, old_version):
```
dralley: Same as above.

pulp_python/app/models.py
```diff
@@ -229,28 +240,39 @@ def _build_additions(self, additions, remote_metadata):
...
         for content_key in removals:
             q |= models.Q(pythonpackagecontent__filename=content_key)
-            q_set = self.repository.content.filter(q)
+            q_set = self.old_version.content().filter(q)
             q_set = q_set.only('id')
             for content in q_set:
                 yield content
 
-    def sync(self):
+    def sync(self, new_version, old_version):
```
dralley: Same as above

pulp_python/app/models.py
```diff
@@ -229,28 +241,39 @@ def _build_additions(self, additions, remote_metadata):
...
+            current_version (pulpcore.plugin.models.RepositoryVersion): the latest pre-existing version
+                or None if one does not exist.
         Returns:
             generator: A generator of content to be removed.
         """
         q = models.Q()
         for content_key in removals:
             q |= models.Q(pythonpackagecontent__filename=content_key)
-            q_set = self.repository.content.filter(q)
+            q_set = self.old_version.content().filter(q)
```
dralley: missed a spot :)

pulp_python/app/models.py
```diff
@@ -229,28 +241,39 @@ def _build_additions(self, additions, remote_metadata):
...
         for content_key in removals:
             q |= models.Q(pythonpackagecontent__filename=content_key)
-            q_set = self.repository.content.filter(q)
+            q_set = current_version.content().filter(q)
             q_set = q_set.only('id')
             for content in q_set:
                 yield content
 
-    def sync(self):
+    def sync(self, new_version, current_version):
```
asmacdo: Sorry to re-nitpick the semantics... This runs in a task, so by the time this runs, it may not be the latest version anymore. I'm not sure how we handle that in pulpcore, but we shouldn't be concerned with whether it is latest or not here. I think "base_version" would be more correct.

pulp_python/app/models.py
```diff
@@ -105,19 +110,23 @@ class PythonImporter(Importer):
...
         Returns:
             set: of contentunit filenames.
         """
         inventory = set()
-
-        q_set = PythonPackageContent.objects.filter(repositories=self.repository)
-        q_set = q_set.only("filename")
-        for content in (c.cast() for c in q_set):
-            inventory.add(content.filename)
+        if current_version is not None:
```
asmacdo: Optional: I'd prefer  to leave this default out of this function. It will be more reusable if it doesn't make assumptions about what to do as if the version isn't specified. I suggest this check should occur as a try/except in the sync code. 

    def sync(self, new_version, base_version):
        try:
            inventory = self._fetch_inventory(base_version)
        except AttributeError: # base_version is None
            inventory = set()

pulp_python/app/models.py
```diff
@@ -170,9 +179,12 @@ def _parse_metadata(cls, project, version, distribution):
         """
         Create a dictionary of metadata needed to create a PythonContentUnit from
         the project, version, and distribution metadata
-        :param project:
-        :param version:
-        :param distribution:
+
+        Args:
```
asmacdo: Optional tweaks:

project (dict): of metadata relevant to the entire Python project
version (string): version of distribution
distribution (dict): of metadata of a single Python distribution

pulp_python/app/models.py
```diff
@@ -105,19 +110,23 @@ class PythonImporter(Importer):
...
     projects = models.TextField(null=True)
 
-    def _fetch_inventory(self):
+    def _fetch_inventory(self, base_version):
         """
         Fetch existing contentunits in the repository.
 
+        Args:
+           base_version (pulpcore.plugin.models.RepositoryVersion): the latest pre-existing
+             version or None if one does not exist.
```
asmacdo: I was thinking s/base_version/version/ since "base" doesn't really have relevance here. 

I think the summary line should be:
"Fetch the content units of the specified version"
and the `version` line should be: "Version to fetch content units from"

pulp_python/app/models.py
```diff
@@ -229,28 +241,39 @@ def _build_additions(self, additions, remote_metadata):
...
             yield content
 
-    def _build_removals(self, removals):
+    def _build_removals(self, removals, base_version):
         """
         Generate the content to be removed.
 
+        Args:
+            removals (set): of filenames to remove
+            base_version (pulpcore.plugin.models.RepositoryVersion): the latest pre-existing
```
asmacdo: same as above

pulp_python/app/models.py
```diff
@@ -229,28 +241,39 @@ def _build_additions(self, additions, remote_metadata):
...
+    def sync(self, new_version, base_version):
         """
         Sync the python projects listed on a :class:`pulp_python.app.models.PythonImporter`
         from a remote repository
+
+        Args:
+           new_version (pulpcore.plugin.models.RepositoryVersion): the new version to which content
+             should be added and removed.
+           base_version (pulpcore.plugin.models.RepositoryVersion): the latest pre-existing
+             version or None if one does not exist.
```
asmacdo: not necessarily "latest".

pulp_python/app/models.py
```diff
@@ -105,24 +110,28 @@ class PythonImporter(Importer):
     TYPE = 'python'
     projects = models.TextField(null=True)
 
-    def _fetch_inventory(self):
+    def _fetch_inventory(self, version):
         """
-        Fetch existing contentunits in the repository.
+        Fetch the contentunits on the specified repository version
```
asmacdo: s/on/in/

PR #125, opened 2018-02-15
Update Python sync to work with core changes

asmacdo:
Changes include:
  Move sync from Importer model to task
  RepositoryVersion context manager
  WorkingDirectory context manager
  natural key default to unique_together

fixes #3294

### DIFF
--- a/pulp_python/app/models.py
+++ b/pulp_python/app/models.py
@@ -1,24 +1,12 @@
-import json
-import os
-
-from collections import namedtuple
 from logging import getLogger
-from urllib.parse import urljoin, urlparse
 
 from django.db import models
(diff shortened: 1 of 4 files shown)


pulp_python/app/models.py
```diff
@@ -79,6 +79,11 @@ class PythonPackageContent(Content):
     obsoletes_dist = models.TextField(default="[]", blank=False)
     requires_external = models.TextField(default="[]", blank=False)
 
+    class Meta:
+        unique_together = (
```
asmacdo: This was added because core either uses a "natural_key" or creates one using "unique_together"

PR #127, opened 2018-02-23
Support multi-resource locking

dralley:
Fix sync support after it was broken by the addition
of multi-resource locking.

re #3186
https://pulp.plan.io/issues/3186


dralley:
mirrors pulp_file PR here: https://github.com/pulp/pulp_file/pull/39

PR #134, opened 2018-03-19
Don't use null=True with TextField

dralley:
closes #3446
https://pulp.plan.io/issues/3446

### DIFF
--- a/pulp_python/app/models.py
+++ b/pulp_python/app/models.py
@@ -44,25 +44,27 @@ class PythonPackageContent(Content):
     """
 
     TYPE = 'python'
+    # Required metadata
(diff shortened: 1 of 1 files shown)


pulp_python/app/models.py
```diff
@@ -110,4 +112,4 @@ class PythonImporter(Importer):
     """
 
     TYPE = 'python'
-    projects = models.TextField(null=True)
+    projects = models.TextField(blank=True)
```
asmacdo: Can this be blank? What circumstance does that make sense?

dralley: You're probably right on that one, I suppose it doesn't make a ton of sense.  Will change.

PR #135, opened 2018-03-20
Support publishing any repository version, not just the latest one

dralley:
* Able to publish any version of the repo, not just the latest one
* Refactoring and cleanup

closes #3504
https://pulp.plan.io/issues/3504

### DIFF
--- a/pulp_python/app/tasks/sync.py
+++ b/pulp_python/app/tasks/sync.py
@@ -11,6 +11,7 @@ from rest_framework import serializers
 
 from pulpcore.plugin import models
 from pulpcore.plugin.changeset import (
+    BatchIterator,
     ChangeSet,
     PendingArtifact,
     PendingContent,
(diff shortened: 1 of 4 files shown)


pulp_python/app/tasks/sync.py
```diff
@@ -173,44 +169,55 @@ def _parse_metadata(project, version, distribution):
...
-        generator: A generator of content to be added.
+        The PythonPackageContent to be added to the repository.
     """
-    for entry in remote_metadata:
-        if(entry['filename'] not in additions):
-            continue
-
-        url = entry.pop('url')
-        artifact = models.Artifact(md5=entry.pop('md5_digest'))
+    def generate():
```
asmacdo: Is there a reason these changes are bundled together?

pulp_python/app/viewsets.py
```diff
@@ -70,18 +70,33 @@ def publish(self, request, pk):
...
+                                  "need to be specified.")
+
+        if 'repository' in request.data and request.data['repository']:
+            repository = self.get_resource(request.data['repository'], Repository)
+
+        if 'repository_version' in request.data and request.data['repository_version']:
+            repository_version = self.get_resource(request.data['repository_version'],
+                                                   RepositoryVersion)
+
+        if repository and repository_version:
```
asmacdo: I'm not convinced this is necessary. If both are specified, why not just use the more specific of the 2, (version). `repository` is still correct.

dralley: I was following what was laid out by the last sentence here: https://pulp.plan.io/issues/3360

pulp_python/app/tasks/sync.py
dralley: Nope.  I can still split them apart but since I had done the refactor first, it was already there, and I didn't notice until after making these changes.

It really isn't a problem to split them up though if you want me to.

asmacdo: Dont worry about it, just curious if there was a connection.

pulp_python/app/viewsets.py
asmacdo: I don't follow, but I can see that there could be some confusion, particularly if the RepositoryVersion.repository doesn't match Respository. Please leave this as you did it.

PR #136, opened 2018-04-05
Rename importer to remote

dralley:


pulp_python/app/tasks/sync.py
```diff
@@ -140,31 +140,31 @@ def _parse_metadata(project, version, distribution):
 
     package = {}
 
-    package['filename'] = distribution['filename']
-    package['packagetype'] = distribution['packagetype']
-    package['name'] = project['name']
+    package['filename'] = distribution.get('filename') or ""
```
werwty: I don't think we should ever allow this to be "", since it's a required field

dralley: You're right but we need a better way to validate that generally.  This patch doesn't actually change that because previously, it was setting the field to ```NULL``` which is also invalid.

pulp_python/app/tasks/sync.py
```diff
@@ -87,11 +87,11 @@ def _fetch_remote(importer):
     """
     remote = []
```
werwty: probably rename this :)

PR #144, opened 2018-04-18
Remove trailing slash from json api.

werwty:
This is caused by the warehouse migrations- the json api works
differently in warehousem and a trailing slash would cause it
to be redirected to a different location.

fixes #3578
https://pulp.plan.io/issues/3578

PR #145, opened 2018-04-19
Fix pypi no more path in json file

Poil:
We also need this fix https://github.com/pulp/pulp_python/pull/144 

ref. https://pulp.plan.io/issues/3578


Poil:
I'm not sure if we must keep the old way (via URL construction)

werwty:
ok test

PR #146, opened 2018-04-20
Add Travis CI

dralley:


.travis/script.sh
```diff
@@ -0,0 +1,29 @@
...
+pulp-manager migrate --noinput
+if [ $? -ne 0 ]; then
+  result=1
+fi
+pulp-manager reset-admin-password --password admin
+pulp-manager runserver >>~/django_runserver.log 2>&1 &
+celery worker -A pulpcore.tasking.celery_app:celery -n resource_manager@%h -Q resource_manager -c 1 --events --umask 18 >>~/resource_manager.log 2>&1 &
+celery worker -A pulpcore.tasking.celery_app:celery -n reserved_resource_worker_1@%h -c 1 --events --umask 18 >>~/reserved_workers-1.log 2>&1 &
+sleep 5
+py.test -v --color=yes --pyargs pulp_smash.tests.pulp3
```
dralley: We need some way to discriminate about which tests should be run here.  Plugins should only run their own tests, I don't think they need to care about core and they definitely shouldn't try to run the tests of other plugins.

Once we get that figured out, the change will need to be mirrored to pulpcore and file.

pulp_python/app/tasks/publish.py
```diff
@@ -145,17 +145,11 @@ def write_simple_api(publication):
 
 def sanitize_name(name):
     """
-    As described by the reference doc, the canonical name for a python distribution is
-    "all lowercase, with dashes replaced by underscores."
+    Strips out all non-alphanumeric characters, including underscores, and replaces them with
+    hyphens. Runs of multiple non-alphanumeric characters are replaced by only one hyphen.
 
-    https://wiki.python.org/moin/PyPISimple
```
asmacdo: Can this be a separate PR please?

PR #147, opened 2018-04-21
Use JSON field from pulpcore instead of trying to store JSON directly in a text field

dralley:
Requires PR https://github.com/PulpQE/pulp-smash/pull/963


dralley:
Just getting the smash tests run atm

PR #150, opened 2018-04-24
Removed sync mode and download policy from remote

dralley:
And docs. And ignore migrations.

closes #3596
https://pulp.plan.io/issues/3596
re #3492
https://pulp.plan.io/issues/3492


.gitignore
```diff
@@ -58,3 +58,8 @@ target/
 
 # PyCharm
 .idea
+
+# Django migrations
```
asmacdo: +1, thank you

pulp_python/app/tasks/sync.py
```diff
@@ -50,8 +50,7 @@ def sync(remote_pk, repository_pk):
             remote_metadata = _fetch_remote(remote)
             remote_keys = set([content['filename'] for content in remote_metadata])
 
-            mirror = remote.sync_mode == 'mirror'
-            delta = _find_delta(inventory=inventory, remote=remote_keys, mirror=mirror)
+            delta = _find_delta(inventory=inventory, remote=remote_keys, mirror=True)
```
asmacdo: i think remove `mirror` and its uses altogether, but this is fine if you prefer.

dralley: I'll do that

PR #153, opened 2018-04-26
Add sync and publish serializers

asmacdo:
https://pulp.plan.io/issues/3597
fixes #3597

### DIFF
--- a/pulp_python/app/serializers.py
+++ b/pulp_python/app/serializers.py
@@ -1,8 +1,9 @@
 from gettext import gettext as _
 
 from rest_framework import serializers
-from pulpcore.plugin import serializers as platform
-from pulpcore.plugin.models import Artifact, ContentArtifact
+from rest_framework_nested.relations import NestedHyperlinkedRelatedField
+from pulpcore.plugin import models as core_models
+from pulpcore.plugin import serializers as core_serializers
 
 from pulp_python.app import models as python_models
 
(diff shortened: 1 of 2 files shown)


daviddavis:
We need these exact same serializers in pulp_ansible. Thoughts on moving them to pulpcore?

daviddavis:
A couple things I noticed when looking at the api schema:

* I don't see your changes in the api schema. I think you need to set the serializer class (e.g. `serializer_class= RepositorySyncURLSerializer`) for the `@detail_route` decorator calls.
* Also, since the api schema now has these definitions based on serializer names (e.g. `RepositorySyncURL`) you need to namespace your serializers so they don't conflict with other plugins. Or you can move these serializers to pulpcore.

asmacdo:
For now, I renamed the serializers and kept them. I filed https://pulp.plan.io/issues/3617 to add sync and publish serializers to core.

daviddavis:
That works. Thanks @asmacdo.

dralley:
@asmacdo if we're merging this here, could you go ahead and include this fix?

pulp_python/app/serializers.py
```diff
@@ -203,5 +204,61 @@ class Meta:
...
+        if not repository and not repository_version:
+            raise serializers.ValidationError(
+                _("Either the 'repository' or 'repository_version' need to be specified"))
+        elif not repository and repository_version:
+            return data
+        elif repository and not repository_version:
+            version = core_models.RepositoryVersion.latest(repository)
+            if version:
+                new_data = {'repository_version': version}
+                new_data.update(data)
```
asmacdo: I'm not a huge fan of altering the data at validation time like this.

PR #154, opened 2018-04-26
Use sha256 hashes intead of md5

dralley:
md5 was the default for old PyPI but Warehouse uses sha256


dralley:
*shakes fist at sqlite*

dralley:
Also updated the mvp page.

PR #155, opened 2018-04-26
Validate that repository has a version to publish

dralley:
Raise a ValidationError if you try to publish a repository
which does not have any versions.

re: 3583
https://pulp.plan.io/issues/3583


dralley:
Actually, I will wait until the serializer thing gets worked out, and then do this a different way.  Don't review this as-is.

PR #156, opened 2018-04-27
Add minimal serializer for PythonPackageContent

dralley:
closes #3624
https://pulp.plan.io/issues/3624

### DIFF
--- a/pulp_python/app/serializers.py
+++ b/pulp_python/app/serializers.py
@@ -223,8 +223,7 @@ class PythonPackageContentSerializer(core_serializers.ContentSerializer):
         return PythonPackageContent
 
     class Meta:
-        fields = (
-            '_href', 'created', 'type',
+        fields = tuple(set(core_serializers.ContentSerializer.Meta.fields) - {'artifacts'}) + (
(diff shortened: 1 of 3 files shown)


dralley:
Please take a look at the story to groom, and then changes will be reflected here.

https://pulp.plan.io/issues/3624

pulp_python/app/serializers.py
```diff
@@ -234,6 +234,18 @@ class Meta:
...
 
+class MinimalPythonPackageContentSerializer(PythonPackageContentSerializer):
+    """
+    A Serializer for PythonPackageContent.
+    """
+
+    class Meta:
+        fields = core_serializers.ContentSerializer.Meta.fields + (
+            'filename', 'packagetype', 'name', 'version'
+        )
```
daviddavis: It looks like PythonPackageContentSerializer uses 'artifact' instead of 'artifacts'. In order to be consistent, what about doing:

```python
fields = tuple(set(core_serializers.ContentSerializer.Meta.fields) - {'artifacts'}) + (
    'filename', 'packagetype', 'name', 'version'
)
```

dralley:
@asmacdo needs a re-review

PR #157, opened 2018-04-29
Create a PythonPackageContent from an Artifact uploaded to Pulp

werwty:
Update content unit creation so it takes only a filename and artifact href

closes #3197, #3604

https://pulp.plan.io/issues/3197
https://pulp.plan.io/issues/3604

### DIFF
--- a/pulp_python/app/tasks/sync.py
+++ b/pulp_python/app/tasks/sync.py
@@ -20,7 +20,7 @@ from pulpcore.plugin.changeset import (
 from pulpcore.plugin.tasking import WorkingDirectory, UserFacingTask
 
 from pulp_python.app import models as python_models
-
+from pulp_python.app.utils import parse_metadata
 
 log = logging.getLogger(__name__)
 
@@ -96,7 +96,7 @@ def _fetch_remote(remote):
         metadata = json.load(open(downloader.path))
         for version, packages in metadata['releases'].items():
             for package in packages:
-                remote_units.append(_parse_metadata(metadata['info'], version, package))
+                remote_units.append(parse_metadata(metadata['info'], version, package))
 
     return remote_units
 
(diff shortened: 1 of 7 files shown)


werwty:
ok test

pulp_python/app/viewsets.py
```diff
@@ -17,6 +41,42 @@ class PythonPackageContentViewSet(platform.ContentViewSet):
...
+        else:
+            raise ValueError(_("Filename extension is not a known python extension."))
+
+        data = parse_project_metadata(meta.__dict__)
+        data['classifiers'] = [{'name': classifier} for classifier in meta.classifiers]
+        data['packagetype'] = meta.packagetype
+        data['version'] = meta.version
+
+        # Update dictionary with user supplied metadata
+        data.update(request.data)
```
werwty: Right now filename definitely need to be user provided in order for pkginfo to be able to parse it, should we make it possible to override other metadata? 
Old pypi use to support override, but warehouse does not...

pulp_python/app/utils.py
```diff
@@ -0,0 +1,64 @@
...
+        dictionary: of useful python metadata
+    """
+
+    package = {}
+
+    package['filename'] = distribution.get('filename') or ""
+    package['packagetype'] = distribution.get('packagetype') or ""
+    package['version'] = version
+    package['url'] = distribution.get('url') or ""
+    package['md5_digest'] = distribution.get('md5_digest') or ""
```
dralley: could you change this to use sha256?

https://github.com/pulp/pulp_python/pull/154

werwty: Is sha256_digest (`distribution['digests']['sha256']`) guaranteed to be present in the json API?

dralley: according to this, it should, but the key is nested under "digests" instead of being at the top level like ```md5_digest``` https://warehouse.pypa.io/api-reference/json/

pulp_python/app/tasks/sync.py
```diff
@@ -121,51 +121,6 @@ def _find_delta(inventory, remote):
     return Delta(additions=additions, removals=removals)
 
 
-def _parse_metadata(project, version, distribution):
```
asmacdo: +1 to move this, thanks

pulp_python/app/viewsets.py
```diff
@@ -17,6 +41,42 @@ class PythonPackageContentViewSet(platform.ContentViewSet):
...
+        for ext, packagetype in DIST_EXTENSIONS.items():
+            if filename.endswith(ext):
+                with tempfile.TemporaryDirectory() as td:
+                    temp_path = os.path.join(td, filename)
+                    shutil.copy2(artifact.file.path, temp_path)
+                    metadata = DIST_TYPES[packagetype](temp_path)
+                    metadata.packagetype = packagetype
+                    break
+        else:
+            raise ValueError(_("Filename extension is not a known python extension."))
```
asmacdo: could you add the filename to this msg?

pulp_python/app/viewsets.py
```diff
@@ -17,6 +41,42 @@ class PythonPackageContentViewSet(platform.ContentViewSet):
...
+
+        data = parse_project_metadata(metadata.__dict__)
```
asmacdo: would `parse_project_metadata(dict(metadata))` work? I think that would be more pythonic.

pulp_python/app/viewsets.py
```diff
@@ -17,6 +41,42 @@ class PythonPackageContentViewSet(platform.ContentViewSet):
...
+        data['classifiers'] = [{'name': classifier} for classifier in metadata.classifiers]
+        data['packagetype'] = metadata.packagetype
+        data['version'] = metadata.version
+
+        # Update dictionary with user supplied metadata
+        data.update(request.data)
+        serializer = self.get_serializer(data=data)
+        serializer.is_valid(raise_exception=True)
+        content = serializer.save()
+        content.artifact = artifact
```
asmacdo: is content.artifact saved?

pulp_python/app/viewsets.py
```diff
@@ -17,6 +41,42 @@ class PythonPackageContentViewSet(platform.ContentViewSet):
...
+
+        filename = request.data['filename']
+
+        # iterate through extensions since splitext does not support things like .tar.gz
+        for ext, packagetype in DIST_EXTENSIONS.items():
+            if filename.endswith(ext):
+                with tempfile.TemporaryDirectory() as td:
+                    temp_path = os.path.join(td, filename)
+                    shutil.copy2(artifact.file.path, temp_path)
+                    metadata = DIST_TYPES[packagetype](temp_path)
```
asmacdo: Its probably worth a comment that the file passed to pkginfo needs the actual filename with extension.

pulp_python/app/viewsets.py
werwty: I don't think you can call `dict()` on a object to get its attributes. But I will use `vars()` instead which looks a lot nicer than using `object.__dict__`

pulp_python/app/viewsets.py
```diff
@@ -17,6 +41,47 @@ class PythonPackageContentViewSet(platform.ContentViewSet):
...
+            raise serializers.ValidationError(_("Extension on {} is not a valid python"
+                                                " extension (.whl, .exe, .egg, .tar.gz, .tar.bz2, "
+                                                ".zip)").format(filename))
+
+        data = parse_project_metadata(vars(metadata))
+        data['classifiers'] = [{'name': classifier} for classifier in metadata.classifiers]
+        data['packagetype'] = metadata.packagetype
+        data['version'] = metadata.version
+        data['filename'] = filename
+        data['artifact'] = request.data['artifact']
```
asmacdo: nice!

PR #158, opened 2018-05-01
Docs reorganization

asmacdo:


docs/workflows/index.rst
```diff
@@ -0,0 +1,39 @@
...
+    password admin
+
+If you configured the ``admin`` user with a different password, adjust the configuration
+accordingly. If you prefer to specify the username and password with each request, please see
+``httpie`` documentation on how to do that.
+
+To make these workflows copy/pastable, we make use of environment variables. The first variable to
+set is the hostname and port.::
+
+   $ export PULP_BASE=http://<hostname>:8000
```
dralley: I suggest that this be named something different.  ```PULP_BASE``` sounds like like it would be the Pulp API root or something, i.e. ```http://<hostname>:8000/pulp/```

Maybe ```BASE_ADDR``` or ```ADDRESS```?  It's just a suggestion so if you disagree, I'm fine with not doing this.

docs/workflows/publish-host.rst
```diff
@@ -0,0 +1,92 @@
...
+
+Create a variable for convenience.::
+
+$ export PUBLICATION_HREF=$(http $PULP_BASE/pulp/api/v3/publications/ | jq -r --arg PUBLISHER_HREF "$PUBLISHER_HREF" '.results[] | select(.publisher==$PUBLISHER_HREF) | ._href')
+
+Host a Publication (Create a Distribution)
+--------------------------------------------
+
+To host a publication, (which makes it consumable by ``pip``), users create a distribution which
+will serve the associated publication in the content app, namespaced by ``base_path``.::
```
dralley: Could this language be simplified or explained a little bit?  A user wouldn't necessarily know what the content app is or means and the term itself probably isn't relevant to them.  They just need to know that it becomes available at ```/pulp/content/```, which we do explain below.

docs/workflows/publish-host.rst
```diff
@@ -0,0 +1,92 @@
...
+If you don't want to specify the distribution path every time, you can modify your ``pip.conf``
+file. See the `pip docs <https://pip.pypa.io/en/stable/user_guide/#configuration>`_ for more
+detail.::
+
+$ cat pip.conf
+
+.. code::
+
+  [global]
+  index-url = http://dev.pulp3:8000/pulp/content/foo/simple/
```
dralley: Don't use a variable specific to our vagrant environment (and I think this change hasn't even happened yet).  If we use ```localhost``` everywhere else we should probably use it here.

docs/workflows/publish-host.rst
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+
+Use the newly created distribution
+-----------------------------------
+
+The metadata and packages can now be retrieved from the distribution::
+
+$ curl localhost:8000/pulp/content/foo/simple/
+$ curl localhost:8000/pulp/content/foo/simple/shelf-reader/
```
dralley: for the sake of consistency, why not use ```$PULP_BASE``` (or whatever the variable name will end up being).

docs/workflows/publish-host.rst
```diff
@@ -0,0 +1,92 @@
...
+
+The content is also pip installable.::
+
+$ pip install --trusted-host dev.pulp3 -i http://dev.pulp3:8000/pulp/content/foo/simple/ shelf-reader
```
dralley: Don't use a hostname specific to our vagrant environment (and I think this change hasn't even happened yet). If we use localhost everywhere else we should probably use it here.

docs/workflows/sync.rst
```diff
@@ -0,0 +1,106 @@
...
+
+Sync repository foo with remote
+-------------------------------
+
+Use the remote object to kick off a synchronize task by specifying the repository to
+sync with. You are telling pulp to fetch content from the remote and add to the repository::
+
+    $ http POST $REMOTE_HREF'sync/' repository=$REPO_HREF
+
+Response::
```
dralley: I don't care so much about this, but it would be nice if we had all of the "Response" sections using ```localhost``` instead of some using ```localhost``` and some using ```dev.pulp3```.

docs/workflows/sync.rst
```diff
@@ -0,0 +1,106 @@
...
+
+Users can populate their repositories with content from an external source like PyPI by syncing
+their repository.
+
+Create a Repository
+-------------------
+
+Start by creating a new repository named "foo"::
+
+    $ http POST http://localhost:8000/pulp/api/v3/repositories/ name=foo
```
dralley: for the sake of consistency, why not use $PULP_BASE (or whatever the variable name will end up being).

docs/workflows/sync.rst
```diff
@@ -0,0 +1,106 @@
...
+Response::
+
+    {
+        "_href": "http://localhost:8000/pulp/api/v3/repositories/e81221c3-9c7a-4681-a435-aa74020753f2/",
+        ...
+    }
+
+If you want to copy/paste your way through the guide, create an environment variable for the repository URI.::
+
+    $ export REPO_HREF=$(http :8000/pulp/api/v3/repositories/ | jq -r '.results[] | select(.name == "foo") | ._href')
```
dralley: for the sake of consistency, why not use $PULP_BASE (or whatever the variable name will end up being).

docs/workflows/upload.rst
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+Create a variable for convenience::
+
+    $ export CONTENT_HREF=$(http $PULP_BASE/pulp/api/v3/content/python/packages/ | jq -r '.results[] | select(.filename == "shelf_reader-0.1-py2-none-any.whl") | ._href')
+
+Add content to a repository
+---------------------------
+
+Once there is a content unit, it can be added and removed to repositories.::
```
dralley: Either:

```added to and removed from repositories``` or

```added and removed from repositories```

File: docs/index.rst [hunk 10]
```diff
@@ -1,23 +1,29 @@
-Pulp Python Support
-===================
+pulp-python Plugin
+==================
+
+The ``pulp_python`` plugin works extends `pulpcore <https://pypi.python.org/pypi/pulpcore/>`__
```
werwty: s/works//

docs/workflows/publish-host.rst
```diff
@@ -0,0 +1,92 @@
...
+       ...
+    }
+
+
+Use the newly created distribution
+-----------------------------------
+
+The metadata and packages can now be retrieved from the distribution::
+
+$ curl localhost:8000/pulp/content/foo/simple/
```
werwty: Why curl and not httpie?

docs/workflows/upload.rst
```diff
@@ -0,0 +1,65 @@
...
+
+Response::
+
+    {
+        "_href": "http://localhost:8000/pulp/api/v3/artifacts/7d39e3f6-535a-4b6e-81e9-c83aa56aa19e/",
+        ...
+    }
+
+
+Create managable content from an artifact
```
werwty: Maybe remove managable? we don't call it manageable content anywhere else..

PR #159, opened 2018-05-01
Add sync_mode parameter to endpoint that deploys sync task

werwty:
closes #3493
https://pulp.plan.io/issues/3493

### DIFF
--- a/pulp_python/app/viewsets.py
+++ b/pulp_python/app/viewsets.py
@@ -156,13 +156,15 @@ class PythonRemoteViewSet(platform.RemoteViewSet):
         )
         serializer.is_valid(raise_exception=True)
         repository = serializer.validated_data.get('repository')
+        mirror = serializer.validated_data.get('mirror')
 
         result = enqueue_with_reservation(
             tasks.sync,
             [repository, remote],
             kwargs={
                 'remote_pk': remote.pk,
-                'repository_pk': repository.pk
+                'repository_pk': repository.pk,
+                'mirror': mirror
             }
         )
         return platform.OperationPostponedResponse(result, request)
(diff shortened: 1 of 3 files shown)


werwty:
Just noticed https://github.com/pulp/pulp_python/pull/153
I will rebase this PR after that is merged...

dralley:
Use tags : P

werwty:
Looks like the revised changeset will have a sync_mode implementation. I will revise this PR to use that once it is merged, for now leaving as WIP

pulp_python/app/viewsets.py
```diff
@@ -100,14 +100,23 @@ def sync(self, request, pk):
         """
         Dispatches a sync task.
         """
+
+        # Get sync_mode, default to additive
+        sync_mode = request.data.get('sync_mode', 'additive').lower()
+
+        if sync_mode not in ['mirror', 'additive']:
```
asmacdo: We should add a ChoiceField to the serializer, which will handle the validation for us. I already added choices as a constant. 
https://github.com/pulp/pulp/blob/3.0-dev/common/pulpcore/common/constants.py#L30-L37

pulp_python/app/viewsets.py
```diff
@@ -156,13 +156,15 @@ def sync(self, request, pk):
         )
         serializer.is_valid(raise_exception=True)
         repository = serializer.validated_data.get('repository')
+        mirror = serializer.validated_data.get('mirror', True)
```
asmacdo: This value defaults to `True` in core, and that serializer is how this option is documented. So I think we shouldn't default here, since our default wouldn't be connected to the schema.
https://github.com/pulp/pulp/blob/cdf6133394d99bbc21a69fae4599d8271d6b344c/pulpcore/pulpcore/app/serializers/repository.py#L142

werwty:
Thanks @CodeHeeler

PR #161, opened 2018-05-04
Add basic filters to pulp_python

werwty:
closes #3626
https://pulp.plan.io/issues/3626

### DIFF
--- a/pulp_python/app/viewsets.py
+++ b/pulp_python/app/viewsets.py
@@ -4,6 +4,7 @@ import tempfile, shutil, os
 from gettext import gettext as _
 
 from django.db import transaction
+from django_filters.rest_framework import filterset
 from pulpcore.plugin import viewsets as platform
 from pulpcore.plugin.models import RepositoryVersion, Artifact
 from rest_framework import decorators, status, serializers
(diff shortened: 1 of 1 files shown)


pulp_python/app/viewsets.py
```diff
@@ -30,6 +31,15 @@
...
 }
 
+class PythonPackageContentFilter(filterset.FilterSet):
+    class Meta:
+        model = python_models.PythonPackageContent
+        fields = {
+            'type': ['exact', 'in'],
+            'author': ['exact', 'in'],
+            'packagetype': ['exact', 'in'],
+            'filename': ['exact', 'in', 'contains'],
```
werwty: Added a contains filter here- this is useful for users trying to find all distributions of a project version:
`http://pulp3.dev:8000/pulp/api/v3/content/python/packages/?filename__contains=pulpcore-3.0.0`

pulp_python/app/viewsets.py
```diff
@@ -82,6 +93,15 @@ def create(self, request):
         return Response(serializer.data, status=status.HTTP_201_CREATED, headers=headers)
 
 
+class PythonRemoteFilter(filterset.FilterSet):
+    class Meta:
+        model = python_models.PythonRemote
+        fields = {
+            'name': ['exact', 'in'],
+            'last_updated': ['lt', 'lte', 'gt', 'gte', 'range'],
+            'projects': ['contains']
```
werwty: I changed this to a 'contains' filter instead of an ['exact', 'in'] filter because that make a lot more sense for a string of projects... It might make sense to add ['in'] if we make this field into a json.

dralley:
Tags :)

werwty:
@dralley Don't know what we'd do without you :)

PR #162, opened 2018-05-04
Keywords string serialized to Char not List

CodeHeeler:
The keywords serializer was serializing a string to a list.
Now it is serializing to a char instead.

fixes #3628
https://pulp.plan.io/issues/3628


werwty:
Could you make sure the default for parsing gets updated to a string? https://github.com/pulp/pulp_python/blob/3.0-dev/pulp_python/app/utils.py#L20

PR #164, opened 2018-05-08
Prepare setup.py for release

werwty:
Update tests to account for smash changes
https://github.com/PulpQE/pulp-smash/pull/993


pulp_python/tests/functional/api_v3/test_publish.py
```diff
@@ -58,14 +58,14 @@ def test_all(self):
                 repo['_versions_href'],
                 {'add_content_units': [file_content['_href']]}
             )
-        versions = get_version_hrefs(repo)
+        versions = get_versions(repo)
         non_latest = choice(versions[:-1])
```
Ichimonji10: This variable should be defined like so:

```python
non_latest = choice(version_hrefs[:-1])
```

PR #168, opened 2018-05-10
Group imports together

Ichimonji10:
See: https://www.python.org/dev/peps/pep-0008/#imports

### DIFF
--- a/pulp_python/tests/functional/api_v3/test_orphans.py
+++ b/pulp_python/tests/functional/api_v3/test_orphans.py
@@ -3,6 +3,7 @@ from unittest import skip
 from random import choice
 
 from pulp_smash import api, cli, config, utils
+from pulp_smash.constants import FILE2_URL
 from pulp_smash.exceptions import CalledProcessError
 from pulp_smash.tests.pulp3.constants import ARTIFACTS_PATH, REPO_PATH
 from pulp_smash.tests.pulp3.utils import (get_auth, get_content, get_versions,  # noqa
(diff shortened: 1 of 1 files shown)


dralley:
Hey Jeremy, thanks for taking a look at these.

I'm going to leave this line the way it is for now, because it's hopefully going to be removed soon, and until then I'd prefer it stuck out as a reminder.

Our current test fixtures aren't good enough to test all of the things that the file plugin tests are testing (tracking issue here: https://pulp.plan.io/issues/3608), so what I've done in the interim is to copy those tests over anyways, disable them, clean them up as much as I could, but leave a few bits of code behind to make it easier to follow when the new fixtures are finished and we want to get these tests enabled again.

Ichimonji10:
Respectfully, I would opine that the way to track work to be done is by using an issue tracker, not by making code messy in ways big or small. This particular technical debt makes code harder to read and linter output harder to parse. Consider the command I had to use when validating https://github.com/pulp/pulp_python/pull/169:

```
pylint \
  --disable=unused-import,missing-docstring,fixme,ungrouped-imports,unsubscriptable-object,duplicate-code \
  --reports=n \
  pulp_python/tests/functional/
```

No fun.

dralley:
That is reasonable, although I'm holding myself personally accountable to make sure it's very short-term technical debt, as I do want to get those tests fully working as soon as the improved fixtures are ready (like, within a week).

As a sidenote, I might look into whether we should use something more strict than the ```flake8``` checking that currently happens on each PR.

dralley:
I would reopen this and go ahead and merge it but for some reason I'm not able to.

Ichimonji10:
Fair enough.

flake8 is good for catching style issues, but it doesn't look for broken imports, incorrect return values, various code smells, and more. We strengthen it by installing several plugins, including flake8-docstrings, flake8-quotes, and pydocstyle. We also use pylint. See the relevant sections of Smash's [setup.py](https://github.com/PulpQE/pulp-smash/blob/master/setup.py#L52-L57) and [Makefile](https://github.com/PulpQE/pulp-smash/blob/master/Makefile#L38-L50).

PR #170, opened 2018-05-11
Mirror Pulp Smash changes

dralley:
https://github.com/PulpQE/pulp-smash/pull/997
https://github.com/PulpQE/pulp-smash/commit/d3987c0662774b1dff6f7500f889211aa05f6f25
https://github.com/PulpQE/pulp-smash/commit/67a0090f9988bcf5ee8b4e6fea330bea90243ef2

Also made a change to the installation docs.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -26,8 +26,8 @@ From Source
 
    sudo -u pulp -i
    source ~/pulpvenv/bin/activate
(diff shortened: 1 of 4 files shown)


dralley:
I realized, *after* doing the work, that there is no real need to duplicate the repository version tests here since there's nothing plugin-specific about them, and the only reason they were in the file test subdirectory is that you have to use *some* content type to test it.

But they pass and the work is done so, why not.

PR #172, opened 2018-05-18
Update auto-distribution tests from pulp-smash

dralley:
Mirror changes from PRs
https://github.com/PulpQE/pulp-smash/pull/1012
https://github.com/PulpQE/pulp-smash/pull/1015


dralley:
Waiting until the second one merges to add that code

PR #173, opened 2018-05-19
As a user, I can whitelist packages to sync with standard python syntax

werwty:
closes #2041
https://pulp.plan.io/issues/2041

### DIFF
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -12,9 +12,7 @@ pip install -r test_requirements.txt
 
 cd .. && git clone https://github.com/pulp/pulp.git
 
-if [ -z $PULP_PR_NUMBER ]; then
-  pushd pulp && git checkout 3.0-dev && popd
-else
+if [ -n $PULP_PR_NUMBER ]; then
   pushd pulp
   git fetch origin +refs/pull/$PULP_PR_NUMBER/merge
   git checkout FETCH_HEAD
--- a/docs/workflows/sync.rst
+++ b/docs/workflows/sync.rst
@@ -32,11 +32,13 @@ itself, a fixture, or even an instance of Pulp 2.
 
 You can use any Python remote to sync content into any repository::
 
-
     $ http POST $BASE_ADDR/pulp/api/v3/remotes/python/ \
         name='bar' \
-        url='https://repos.fedorapeople.org/repos/pulp/pulp/fixtures/python-pypi/' \
-        projects='["shelf-reader"]'
+        url='https://pypi.org/' \
+        projects:='[{"name": "django", "version_specifier":"~=2.0"}]'
+
+
+
 
 Response::
 
@@ -48,6 +50,36 @@ Response::
 Again, you can create an environment variable for convenience::
 
     $ export REMOTE_HREF=$(http $BASE_ADDR/pulp/api/v3/remotes/python/ | jq -r '.results[] | select(.name == "bar") | ._href')
+
+
+A More Complex Remote
+---------------------
+If only the name of a project is specified, every distribution of every version of that project
+will be synced. You can use the version_specifier and digest fields on a project to ensure
+only distributions you care about will be synced::
+
+    $ http POST $BASE_ADDR/pulp/api/v3/remotes/python/ \
+        name='complex-remote' \
+        url='https://pypi.org/' \
+        projects:='[
+            { "name": "django",
+              "version_specifier": "~=2.0,!=2.0.1",
+              "digests":[
+                    {"type": "sha256",
+                     "digest": "3d9916515599f757043c690ae2b5ea28666afa09779636351da505396cbb2f19"}
+              ]
+            },
+            {"name": "pip-tools",
+             "version_specifier": ">=1.12,<=2.0"},
+            {"name": "scipy",
+             "digests":[
+                {"type": "md5",
+                "digest": "044af71389ac2ad3d3ece24d0baf4c07"},
+                {"type": "sha256",
+                "digest": "18b572502ce0b17e3b4bfe50dcaea414a98290358a2fa080c36066ba0651ec14"}]
+            },
+            {"name": "shelf-reader"}
+        ]'
 
 
 Sync repository foo with remote
--- a/pulp_python/app/models.py
+++ b/pulp_python/app/models.py
@@ -34,6 +34,47 @@ class Classifier(Model):
                                                on_delete=models.CASCADE)
 
 
+class DistributionDigest(Model):
+    """
+    A model of digests on an individual distribution.
+    """
+    type = models.TextField()
+    digest = models.TextField()
+    project_specifier = models.ForeignKey("ProjectSpecifier",
+                                          related_name="digests",
+                                          related_query_name="distributiondigest",
+                                          on_delete=models.CASCADE)
+
+
+class ProjectSpecifier(Model):
+    """
+    A specifier of a python project.
+
+    Example:
+
+        digests: ["sha256:0000"] will only match the distributions that has the exact hash
+        name: "projectname" without specifiers will match every distribution in the project.
+        version_specifier: "==1.0.0" will match all distributions matching version
+        version_specifier: "~=1.0.0" will match all major version 1 distributions
+        version_specifier: "==1.0.0" digests: ["sha256:0000"] will only match the distributions
+            with the hash and with version 1.0.0
+        version_specifier: ">=0.9,<1.0" will match all versions matching 0.9.*
+
+    Fields:
+
+        name (models.TextField): The name of a python project
+        version_specifier (models.TextField):  Used to filter the versions of a project to sync
+        remote (models.ForeignKey): The remote this project specifier is associated with
+    """
+
+    name = models.TextField()
+    version_specifier = models.TextField(blank=True, default="")
+    remote = models.ForeignKey("PythonRemote",
+                               related_name="projects",
+                               related_query_name="projectspecifier",
+                               on_delete=models.CASCADE)
+
+
 class PythonPackageContent(Content):
     """
     A Content Type representing Python's Distribution Package as
@@ -104,4 +145,3 @@ class PythonRemote(Remote):
     """
 
     TYPE = 'python'
-    projects = models.TextField()
--- a/pulp_python/app/serializers.py
+++ b/pulp_python/app/serializers.py
@@ -1,9 +1,11 @@
 from gettext import gettext as _
 
-from rest_framework import serializers
-from rest_framework_nested.relations import NestedHyperlinkedRelatedField
+from django.db import transaction
+from packaging import specifiers
 from pulpcore.plugin import models as core_models
 from pulpcore.plugin import serializers as core_serializers
+from rest_framework import serializers
+from rest_framework_nested.relations import NestedHyperlinkedRelatedField
 
 from pulp_python.app import models as python_models
 
@@ -20,6 +22,63 @@ class ClassifierSerializer(serializers.ModelSerializer):
     class Meta:
         model = python_models.Classifier
         fields = ('name',)
+
+
+class DistributionDigestSerializer(serializers.ModelSerializer):
+    """
+    A serializer for the Distribution Digest in a Project Specifier
+    """
+    type = serializers.CharField(
+        help_text=_("A type of digest: i.e. sha256, md5")
+    )
+
+    digest = serializers.CharField(
+        help_text=_("The digest of the distribution")
+    )
+
+    class Meta:
+        model = python_models.DistributionDigest
+        fields = ('type', 'digest')
+
+
+class ProjectSpecifierSerializer(serializers.ModelSerializer):
+    """
+    A serializer for Python project specifiers
+    """
+
+    name = serializers.CharField(
+        help_text=_("A python project name.")
+    )
+
+    version_specifier = serializers.CharField(
+        help_text=_("A version specifier, accepts standard python versions "
+                    "syntax: >=, <=, ==, ~=, >, <, ! can be used in conjunction with"
+                    "other specifiers i.e. >1,<=3,!=3.0.2. Note that the specifiers "
+                    "treat pre-released versions as < released versions, so 3.0.0a1 < 3.0.0."
+                    "Not setting the version_specifier will sync all the pre-released and"
+                    "released versions."),
+        required=False,
+        allow_blank=True
+    )
+
+    digests = DistributionDigestSerializer(
+        required=False,
+        many=True
+    )
+
+    def validate_version_specifier(self, value):
+        """
+        Check that the Version Specifier is valid
+        """
+        try:
+            specifiers.SpecifierSet(value)
+        except specifiers.InvalidSpecifier as err:
+            raise serializers.ValidationError(err)
+        return value
+
+    class Meta:
+        model = python_models.ProjectSpecifier
+        fields = ('name', 'version_specifier', 'digests')
 
 
 class PythonPackageContentSerializer(core_serializers.ContentSerializer):
@@ -141,10 +200,14 @@ class PythonPackageContentSerializer(core_serializers.ContentSerializer):
     def create(self, validated_data):
         """
         Creates a PythonPackageContent
-
         Overriding default create() to write the classifiers nested field
-        :param validated_data:
-        :return:
+
+         args:
+            validated_data (dict): data used to create the PythonPackageContent
+
+        returns:
+            models.PythonPackageContent: the created PythonPackageContent
+
         """
 
         classifiers = validated_data.pop('classifiers')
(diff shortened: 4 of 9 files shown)


setup.py
```diff
@@ -4,7 +4,8 @@
 
 requirements = [
     'pulpcore-plugin',
-    'pkginfo'
+    'pkginfo',
+    'packaging'
```
asmacdo: Optional: it might be good to add a `,` so future diffs will only be 1 line

pulp_python/app/serializers.py
```diff
@@ -21,6 +21,31 @@ class Meta:
...
+        help_text=_("versionspecifier"),
+        required = False,
+        allow_blank=True
+    )
+
+    sha256 = serializers.CharField(
+        help_text=_("hash"),
+        required=False,
+    )
+
```
asmacdo: If we add a `create` function, we could accept strings in the `requirements.txt` syntax instead a dictionary.

dralley:
Tags por favor :)

pulp_python/app/serializers.py
```diff
@@ -21,6 +21,31 @@ class Meta:
         model = python_models.Classifier
         fields = ('name',)
 
+class ProjectSpecifierSerializer(serializers.ModelSerializer):
+    """
+    A serializer for Python Classifiers
```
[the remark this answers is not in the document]
werwty: s/version/project/

pulp_python/app/serializers.py
```diff
@@ -21,6 +21,31 @@ class Meta:
...
+    """
+
+    name = serializers.CharField(
+        help_text=_("A python project name.")
+    )
+
+    version = serializers.CharField(
+        help_text=_("versionspecifier"),
```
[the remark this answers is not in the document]
werwty: I'm updating this to be
`A version specifier, accepts standard python versions syntax: >=, <=, ==, ~=, >, <`

pulp_python/app/models.py
```diff
@@ -33,6 +33,31 @@ class Classifier(Model):
...
+    Fields:
+
+        project_name (models.TextField): The name of a python project
+        version (modles.TextField): A filter for the versions to sync
+        sha256 (models.CharField): A hash of a distribution to sync
+        remote (models.ForeignKey): The remote this project specifier is associated with
+    """
+
+    name = models.TextField()
+    version = models.TextField(blank=True, default="")
```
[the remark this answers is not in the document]
werwty: yes! packaging.specifiers.SpecifierSet supports syntax like: `">=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*"`
I will call this out on the documentation

docs/workflows/sync.rst
```diff
@@ -35,8 +35,10 @@ You can use any Python remote to sync content into any repository::
 
     $ http POST $BASE_ADDR/pulp/api/v3/remotes/python/ \
         name='bar' \
-        url='https://repos.fedorapeople.org/repos/pulp/pulp/fixtures/python-pypi/' \
-        projects='["shelf-reader"]'
+        url='https://pypi.org/' \
+        projects:='[{"name": "django", "version":">2.0",
+        "sha256":"3d9916515599f757043c690ae2b5ea28666afa09779636351da505396cbb2f19"},
```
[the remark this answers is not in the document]
werwty: I like having a more complex example that shows the users what they can do without having to read up a separate section of docs.

dralley:
This looks great @werwty , I'll test it tomorrow

dralley:
@pep8speaks

werwty:
@dralley hold off for a day please, I still need to refactor sync and update it based on asmacdo's other comments.

asmacdo:
As much as I'd like to merge, I think we really have to wait until https://pulp.plan.io/issues/3705 is fixed.

pulp_python/app/serializers.py
```diff
@@ -22,6 +23,50 @@ class Meta:
...
+class ProjectSpecifierSerializer(serializers.ModelSerializer):
+    """
+    A serializer for Python project specifiers
+    """
+
+    name = serializers.CharField(
+        help_text=_("A python project name.")
+    )
+
+    version_specifier = serializers.CharField(
```
dralley: It would be good to add validation to the version specifiers, although it may be non-trivial enough to make a future task.

http://www.django-rest-framework.org/api-guide/serializers/#field-level-validation

pulp_python/app/models.py
```diff
@@ -34,6 +34,45 @@ class Classifier(Model):
...
+class ProjectSpecifier(Model):
+    """
+    A specifier of a python project dependency.
+
+    Example:
+
+        version==1.0.0 will sync all distributions matching version
+        version~=1.0.0 will sync all major version 1 distributions
+        version==1.0.0 digests=["sha256:0000"] will only sync the distributions matching the hash
+        name=projectname without specifiers will sync every distribution in the project.
```
dralley: Does digests work separately from a version specifier?  If yes, could you provide an example for just digests?

Could you provide another example using  ```>=0.9, <1.0```?

pulp_python/app/serializers.py
```diff
@@ -22,6 +23,50 @@ class Meta:
...
+        help_text=_("A version specifier, accepts standard python versions "
+                    "syntax: >=, <=, ==, ~=, >, <, can be used in conjunction with"
+                    "multiple specifiers like "),
+        required=False,
+        allow_blank=True
+    )
+
+    digests = DistributionDigestSerializer(
```
dralley: Same here I think, we can/should validate that the digest type is supported and the correct length for whatever type it is.

pulp_python/app/serializers.py
```diff
@@ -22,6 +23,50 @@ class Meta:
         fields = ('name',)
 
 
+class DistributionDigestSerializer(serializers.ModelSerializer):
```
dralley: Same here, we should validate that the digest type is supported and that the digest is the correct length for whatever type it is.

werwty: I'm not sure about this, we really "support" any digest type. During sync time all we do is check that the type and value in the JSON API's digest field is the same.

dralley: Good point, never mind then.

docs/workflows/sync.rst
```diff
@@ -35,8 +35,10 @@ You can use any Python remote to sync content into any repository::
 
     $ http POST $BASE_ADDR/pulp/api/v3/remotes/python/ \
         name='bar' \
-        url='https://repos.fedorapeople.org/repos/pulp/pulp/fixtures/python-pypi/' \
-        projects='["shelf-reader"]'
+        url='https://pypi.org/' \
+        projects:='[{"name": "django", "version_specifier":"~=2.0",
```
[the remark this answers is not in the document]
werwty: nope, it's already enclosed in a `'` so bash knows to wait for it to close

pulp_python/app/serializers.py
```diff
@@ -22,6 +23,50 @@ class Meta:
...
+    """
+    A serializer for the Distribution Digest in a Project Specifier
+    """
+    type = serializers.CharField(
```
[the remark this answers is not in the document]
werwty: I think not combined is nicer, and we don't need to split when comparing hashes

pulp_python/app/serializers.py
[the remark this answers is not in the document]
werwty: I don't think it's all that unusual: if you want a single distribution, just give the hash, why also specify a version?

pulp_python/app/tasks/sync.py
```diff
@@ -75,27 +78,42 @@ def _fetch_inventory(version):
     return inventory
 
 
-def _fetch_remote(remote):
+def _fetch_specified_remotes(remote, project_specifiers):
```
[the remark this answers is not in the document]
werwty: But this only fetches some remote metadata, maybe fetch_specified_remote_metadata? (naming is hard)

pulp_python/app/models.py
```diff
@@ -34,6 +34,45 @@ class Classifier(Model):
...
+
+    Fields:
+
+        project_name (models.TextField): The name of a python project
+        version (modles.TextField): A filter for the versions to sync
```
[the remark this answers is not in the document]
dralley: also typo:  s/modles/models

pulp_python/app/tasks/sync.py
```diff
@@ -75,27 +78,42 @@ def _fetch_inventory(version):
...
         metadata = json.load(open(downloader.path))
         for version, packages in metadata['releases'].items():
-            for package in packages:
-                remote_units.append(parse_metadata(metadata['info'], version, package))
-
+            specifier = specifiers.SpecifierSet(project.version_specifier)
+            if specifier.contains(version):
+                for package in packages:
+                    # add the package if the project specifier does not have an associated digest
+                    if digests.count() == 0:
```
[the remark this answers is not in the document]
werwty: nope, since digests is a QuerySet that doesn't evaluate to false

pulp_python/app/serializers.py
```diff
@@ -141,10 +186,7 @@ class PythonPackageContentSerializer(core_serializers.ContentSerializer):
     def create(self, validated_data):
         """
         Creates a PythonPackageContent
```
dralley: newline + args

pulp_python/app/tasks/sync.py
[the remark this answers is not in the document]
werwty: I can just use exists()

docs/workflows/sync.rst
```diff
@@ -35,8 +35,10 @@ You can use any Python remote to sync content into any repository::
 
     $ http POST $BASE_ADDR/pulp/api/v3/remotes/python/ \
         name='bar' \
-        url='https://repos.fedorapeople.org/repos/pulp/pulp/fixtures/python-pypi/' \
-        projects='["shelf-reader"]'
+        url='https://pypi.org/' \
```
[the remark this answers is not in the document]
dralley: I'm not sure this is actually necessary, or would even work.  The JSON spec doesn't allow a trailing comma IIRC.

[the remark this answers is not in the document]
asmacdo: no i just saw the other comma and assumed this needed one as well.

pulp_python/app/serializers.py
```diff
@@ -178,15 +220,63 @@ class PythonRemoteSerializer(core_serializers.RemoteSerializer):
...
+
+        for project in projects:
+            digests = project.pop('digests', None)
+            specifier = python_models.ProjectSpecifier.objects.create(remote=python_remote,
+                                                                      **project)
+            if digests:
+                for digest in digests:
+                    python_models.DistributionDigest.objects.create(project_specifier=specifier,
+                                                                    **digest)
+
```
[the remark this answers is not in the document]
werwty: digests are cascade deleted with project specifier. 
(and if project specifier is updated the old one gets deleted)

pulp_python/app/serializers.py
```diff
@@ -178,15 +220,63 @@ class PythonRemoteSerializer(core_serializers.RemoteSerializer):
...
+        """
+
+        projects = validated_data.pop('projects', [])
+
+        python_remote = python_models.PythonRemote.objects.get(pk=instance.pk)
+
+        # Remove all project specifier related by foreign key to the remote if it is not a
+        # partial update or if new projects list has been passed
+        if not partial or projects:
+            python_models.ProjectSpecifier.objects.filter(remote=python_remote).delete()
```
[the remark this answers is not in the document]
werwty: They should get cascade deleted with project specifier

pulp_python/app/serializers.py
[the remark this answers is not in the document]
asmacdo: cool. how about a choice field then? Re: https://github.com/pulp/pulp_python/pull/173/files#r192219408, I think we probably do want to only support certain digests. For the moment, we only use digests to filter the content that we download, but it would be good form to check that hash after we download it.

pulp_python/app/models.py
```diff
@@ -34,6 +34,45 @@ class Classifier(Model):
...
+    """
+    type = models.TextField()
+    digest = models.TextField()
+    project_specifier = models.ForeignKey("ProjectSpecifier",
+                                          related_name="digests",
+                                          related_query_name="distributiondigest",
+                                          on_delete=models.CASCADE)
+
+
+class ProjectSpecifier(Model):
```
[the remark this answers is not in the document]
werwty: We've been using the project/version/distribution names throughout. Why change this to package now?

pulp_python/app/tasks/sync.py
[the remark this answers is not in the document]
asmacdo: That works for me. or `fetch_sepecified_metadata`. Its sort of obvious that is going to come from the remote.

pulp_python/app/serializers.py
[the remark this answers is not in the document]
werwty: This digest is user provided and only used to filter. The hash checking logic would not belong on this but rather on the digest of the created artifact

pulp_python/app/models.py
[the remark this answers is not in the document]
asmacdo: Hmm. DistributionSpecifier could be confused with our Distribution, and PackageSpecifier is clear enough. The point is it doesn't specify projects, even though that is how they are organized. They actually specify a set of Python Distributions.

dralley:
Make sure all the smash tests pass -- it may require modifying some of them.

Rebase if you haven't already as I fixed an issue causing some smash tests to fail earlier.

.travis/install.sh
```diff
@@ -37,5 +37,5 @@ else
   popd
 fi
 
-cd pulp_python
```
dralley: This should be incorrect, after entering the directories above it does a popd, so it should be ending up where it started.

Is that not happening properly?

werwty: It is not: https://travis-ci.org/pulp/pulp_python/jobs/386389531

werwty: ohhh I see this is after we renamed the branch so the branch checkout errors out and it doesn't popd

dralley: @werwty this is why

https://travis-ci.org/pulp/pulp_python/jobs/386389531#L550

The branch 3.0-dev no longer exists, therefore the ```git checkout``` errors, therefore bash's ```&&``` expression falls through early.

dralley: @werwty it does still need the ```cd pulp_python``` line though

Also could you edit the install docs?

.travis/install.sh
```diff
@@ -13,7 +13,7 @@ pip install -r test_requirements.txt
 cd .. && git clone https://github.com/pulp/pulp.git
 
 if [ -z $PULP_PR_NUMBER ]; then
-  pushd pulp && git checkout 3.0-dev && popd
+  pushd pulp && git checkout master && popd
```
dralley: Probably no need to have this branch anymore since it defaults to master.

pulp_python/app/models.py
dralley: ^

werwty: @dralley Ah I thought you meant add an example to the workflow docs. I'll add those to the docstring too

pulp_python/app/models.py
```diff
@@ -34,6 +34,47 @@ class Classifier(Model):
...
+    """
+    A specifier of a python project.
+
+    Example:
+
+        version_specifier==1.0.0 will sync all distributions matching version
```
asmacdo: I still think thes examples could be misleading. 

Also, s/sync/match/ throughout this docstring

pulp_python/app/models.py
```diff
@@ -34,6 +34,47 @@ class Classifier(Model):
...
+        version_specifier~=1.0.0 will sync all major version 1 distributions
+        version_specifier==1.0.0 digests=["sha256:0000"] will only sync the distributions
+            matching the hash with version 1.0.0
+        name=projectname without specifiers will sync every distribution in the project.
+        version_specifier>=0.9,<1.0 will sync all versions matching 0.9.*
+        digests=["sha256:0000"] will only sync the distributions matching
+
+    Fields:
+
+        project_name (models.TextField): The name of a python project
```
asmacdo: s/project_name/name/

pulp_python/app/serializers.py
```diff
@@ -1,7 +1,10 @@
 from gettext import gettext as _
 
+from django.db import transaction
+from packaging import specifiers
 from rest_framework import serializers
 from rest_framework_nested.relations import NestedHyperlinkedRelatedField
+
```
asmacdo: no space here. pulpcore is a 3rd party library from the plugin's perspective.

pulp_python/app/serializers.py
```diff
@@ -22,6 +25,60 @@ class Meta:
...
+    )
+
+    digests = DistributionDigestSerializer(
+        required=False,
+        many=True
+    )
+
+    def validate_version_specifier(self, value):
+        """
+        Check that the Version Specifier valid
```
asmacdo: s/valid/is valid/

Nice!

pulp_python/app/tasks/sync.py
```diff
@@ -75,27 +78,51 @@ def _fetch_inventory(version):
...
-                remote_units.append(parse_metadata(metadata['info'], version, package))
-
+                # If neither specifiers nor digests have been set, then we should add the unit
+                if not project.version_specifier and not digests.exists():
+                    remote_units.append(parse_metadata(metadata['info'], version, package))
+                    continue
+
+                specifier = specifiers.SpecifierSet(project.version_specifier)
+
+                # Note: SpecifierSet("").contains(version) will return true for released versions
```
asmacdo: Could you add this to the help_text on the serializer?

pulp_python/app/models.py
werwty: In what way are these examples misleading? (How can I clarify them?)

asmacdo: `version_specifier==1.0.0` is not a valid version_specifier
I think even just adding punctuation will clarify 

"version_specifier: `==1.0.0` will match all distributions ..."

PR #183, opened 2018-06-12
normalize project name before publish

werwty:
closes #3632
https://pulp.plan.io/issues/3632

### DIFF
--- a/plugins/pulp_python/plugins/utils.py
+++ b/plugins/pulp_python/plugins/utils.py
@@ -1,0 +1,15 @@
+import re
+
+
+def sanitize_name(name):
+    """
+    Strips out all non-alphanumeric characters, including underscores, and replaces them with
+    hyphens. Runs of multiple non-alphanumeric characters are replaced by only one hyphen.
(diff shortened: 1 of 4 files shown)


plugins/pulp_python/plugins/utils.py
```diff
@@ -0,0 +1,5 @@
+import re
+
+
+def normalize_project_name(name):
+    return re.sub(r"[-_.]+", "-", name).lower()
```
dralley: I would prefer we just used the same function as with Pulp 3, unless there is a reason otherwise.

https://github.com/pulp/pulp_python/blob/3.0-dev/pulp_python/app/tasks/publish.py#L145

werwty:
ok test

werwty:
Unit test failures are due to a known issue caused by different versions of installed pulp and unittests (one from PR the other from master). Unfortunately this is a intermittent issue and previous attempts at fixing it have broken other things: https://github.com/pulp/pulp-ci/pull/525

I tested this PR locally and all unit tests pass.

PR #190, opened 2018-07-08
Fix OpenAPI schema for async responses

werwty:
closes #3668
https://pulp.plan.io/issues/3668

### DIFF
--- a/pulp_python/app/viewsets.py
+++ b/pulp_python/app/viewsets.py
@@ -1,17 +1,19 @@
 import os
 from gettext import gettext as _
+import pkginfo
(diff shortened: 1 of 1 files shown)


werwty:
ok test

dralley:
@werwty There's no need to run the Pulp 2 Jenkins suite. They ran prior to this https://github.com/pulp/pulp-ci/pull/550 PR but since it's now fixed they aren't going to run again.  That they ran in the first place was a mistake.

werwty:
ack thanks @dralley. Merging as is then

PR #201, opened 2018-08-21
Add exclude support (aka 'blacklist')

dralley:
Renames the "projects" field on the remote to "includes".

Adds a new "excludes" field to the remote which behaves like "includes", except that any specified releasees or digests are *not* synced, even if an include specifier matches them.

Also adds a 'prereleases' field to the remote, which toggles whether prerelease versions should be synced.  This mirrors the 'prereleases' flag that ```packaging.specifiers.SpecifierSet``` provides.

Adds smash tests for the following:

* Test CRUD remote w/ include and exclude
   * valid / invalid include exclude
* Test prereleases true/false
* Test project specifiers include/exclude
    * test adding new includes later increases the count
    * test adding new excludes later decreases the count
    * test removing includes later decreases the count
    * test removing excludes later increases the count


pulp_python/app/tasks/publish.py
```diff
@@ -93,7 +94,7 @@ def write_simple_api(publication):
         .distinct()
     )
 
-    index_names = [(name, sanitize_name(name)) for name in project_names]
+    index_names = [(name, canonicalize_name(name)) for name in project_names]
```
CodeHeeler: I like de-duplication and using canonicalize_name found in packaging.utils instead of importing re and writing one ourselves, but correct me if I'm wrong (my regex is a bit rusty) the sanitize_name stripped out *all* non-alphanumeric characters whereas canonicalize_name only strips out '-', '_', and '.'.  Can we assume we won't encounter any other characters for our purposes?  I know there aren't supposed to be for naming conventions but does PYPI and anything else we may use enforce this?

pulp_python/app/tasks/sync.py
```diff
@@ -82,19 +83,37 @@ def __init__(self, remote):
             out_q (asyncio.Queue): The out_q to send `DeclarativeContent` objects to.
 
         """
-        project_specifiers = ProjectSpecifier.objects.filter(remote=self.remote)
+        ps = ProjectSpecifier.objects.filter(remote=self.remote)
```
CodeHeeler: I am uncomfortable with using the name ps so close to the name pb in the code.  The short names are convenient for line length, but could get mixed up making it less readable which doesn't feel as pythonic as a descriptive name here.

pulp_python/app/tasks/publish.py
dralley: I checked the source code when changing this and it's actually the exact same regex as the current one. ```packaging``` is an official PyPA package and so it makes sense to just use their function instead of having our own copy.

pulp_python/app/tasks/sync.py
dralley: That's reasonable.  The reason it was changed wasn't for brevity though, it's because there are other places below that use the variable name  ```project_specifiers``` and I didn't want them to collide.  So we'd have to come up with another name to use here.

dralley:
re: separate PRs, I agree in principle, but in the process of adding all the smash tests I found and fixed enough bugs that I don't think we would want to just revert this PR in the future.

dralley:
re: scope, I agree in principle, but given how busy everyone is with higher priority projects it's a lot easier to roll them together for now.  Plus, with all the test additions, it would probably not be a good idea to just roll this PR back.

PR #204, opened 2018-09-25
Docs and code use int IDs

vdusek:
closes: #3860
https://pulp.plan.io/issues/3860


vdusek:
Not sure about this - [pulp_python/tests/functional/api/test_crud_remotes.py#L145](https://github.com/pulp/pulp_python/blob/master/pulp_python/tests/functional/api/test_crud_remotes.py#L145), should it be replaced just for a random int (`random.randint()`)?

PR #207, opened 2018-10-30
Add line break to package index page

byoungb:
Just adding the BR tag to the `simple_index_template`

### DIFF
--- a/pulp_python/app/tasks/publish.py
+++ b/pulp_python/app/tasks/publish.py
@@ -22,7 +22,7 @@ simple_index_template = """<!DOCTYPE html>
   </head>
   <body>
     {% for name, canonical_name in projects %}
-    <a href="{{ canonical_name }}">{{ name }}</a>
+    <a href="{{ canonical_name }}">{{ name }}</a><br/>
     {% endfor %}
   </body>
 </html>


dralley:
Hello @byoungb, thanks for taking a look at Pulp!  

I do think adding breaks would make the page more readable, and I don't think it would hurt anything,  but I'm a little conflicted because it's not what PyPI does and we want to emulate PyPI to the greatest extent possible.

https://pypi.org/simple/

I think the index pages *did* use newlines before the transition to warehouse, in which case it would probably be OK to do it, but we'd prefer not to diverge from how PyPI currently does things unless there's a good reason to do so. It's just difficult to know if there's some code out there that it might cause issues for.

byoungb:
I didn't realize that but understand if that is what you were going for then yeah ignore this PR, I am actually surprised that they even have that index page on pypi.org with over 150,000+ packages, it is a pretty unusable page, and as far as I know it is not used by pip or easyinstall.   

I have used to other PyPI servers in the past, and both of them had a richer /simple/ index pages with new lines.   
screenshot from 2018-10-30 16-04-19

Here are the 2 other pypi servers that I have used in the past
https://github.com/pypiserver/pypiserver/blob/cd7707379a35fb4f1e1c2081697a5a02b4a36e0b/pypiserver/_app.py#L241
https://github.com/benliles/djangopypi/blob/master/djangopypi/templates/djangopypi/package_list_simple.html#L7

I did find another bug also with this page... the package names URLs need to end in a / if they do not the relative URLs on the next page go up one level 2 far. I will issue another PR when I am sure I am not going to find anything else.

I am just now switching to Pulp so that I can host RPMs and Ansible Roles also. I really love what I see with Pulp, was a bit hard to wrap my head around at first with the many many layers. But I will probably want to help contribute a bunch more in the future.

byoungb:
Also where is the correct place for me to be putting any ideas I have that I would like to discuss and contribute to for this project and for pulpcore?

dralley:
@byoungb Interesting...  I think it's probably fine then.  It looks like Pulp 2 also still has the </br> tag and we haven't seen any related issues there. @asmacdo, what do you think?

>I am just now switching to Pulp so that I can host RPMs and Ansible  Roles also. I really love what I see with Pulp, was a bit hard to wrap  my head around at first with the many many layers. But I will probably  want to help contribute a bunch more in the future.

That would be great :)  

Our issue tracker is located here: https://pulp.plan.io/

Day-to-day dev chat happens on the #pulp-dev IRC channel on freenode.  Feel free to drop in any time, we would love to hear your feedback / ideas / questions.

asmacdo:
I can speak a little about the simple index. AFAIK it is most used by bandersnatch and puppet (and pulp)--not usually humans. @byoungb if you are hoping to have human readable simple API, I think that's reasonable. I prefer to default to doing the PyPI way (@dralley and I agreed earlier). However if you do need human readable, I doubt this would cause a problem. I bet you could talk us into it :)

I'm not sure if this was a deliberate change in warehouse, but imo worth filing an issue over there. If they are into it, this would be a great first time contribution to Warehouse. (I don't want to, I'll volunteer).

asmacdo:
@byoungb, by the way if you encountered anything... suboptimal in the docs, let us know. We are preparing a docs focused period that will start very soon.

dralley:
I'm just going to merge this, I don't think it hurts anything.

PR #215, opened 2018-12-19
Prepend Content model hierarchy with _

CodeHeeler:
Prepend all pulpcore model fields in the Content model hierarchy with _
and do the same for all plugins reliant on them.

Required PR: https://github.com/pulp/pulp/pull/3798

re #4206
https://pulp.plan.io/issues/4206

### DIFF
--- a/pulp_python/app/models.py
+++ b/pulp_python/app/models.py
@@ -140,7 +140,7 @@ class PythonPackageContent(Content):
         """
         Return the artifact id (there is only one for this content type).
         """
-        return self.artifacts.get().pk
+        return self._artifacts.get().pk
 
     def __str__(self):
         """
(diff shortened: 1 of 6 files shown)


dralley:
Bump the pulpcore-plugin dep to b15, it should fix this issue.

pulp_python/app/models.py
```diff
@@ -152,11 +152,11 @@ def __str__(self):
         e.g. <PythonPackageContent: shelf-reader [version] (whl)>
 
         """
-        return '<{obj_name}: {name} [{version}] ({type})>'.format(
+        return '<{obj_name}: {name} [{version}] ({_type})>'.format(
```
dralley: This actually doesn't need to be changed, it maps to packagetype not the underlying "_type" field.

dralley:
@CodeHeeler @asmacdo , Test failures are due to an upstreah aiohttp issue, not anything we're doing.  I downgraded aiohttp locally and everything passes now.  Merging.

https://github.com/aio-libs/aiohttp/issues/3477

PR #218, opened 2019-01-25
updated the documentation link. fixes #4154

lina-nikiforova:
The readme pointed to https://pulp-python.readthedocs.io/en/3.0-dev/

This versions includes a note: "You are not reading the most recent version of this documentation. 3.0.0b1 is the latest version available."

So I updated the link to the latest release.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -10,5 +10,5 @@ pulp_python
 A Pulp plugin to support hosting your own pip compatible Python packages.
 
 For more information, please see the `documentation
-<http://pulp-python.readthedocs.io/en/3.0-dev/>`_ or the `Pulp project page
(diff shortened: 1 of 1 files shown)


asmacdo:
Hey @lina-nikiforova! Thanks for this :)

I finally had a chance to look at this. This PR is an improvement, but it would be even better if the link pointed to https://pulp-python.readthedocs.io/en/latest/. Would you be willing to make that change?

lina-nikiforova:
@asmacdo sure thing, it makes sense :)

dralley:
Lina, you're right that the test failure is totally unrelated - it's
expected from a change we made to Pulp last week.  I'm planning to fix it
today, but we can ignore the failure for the purpose of this docs PR.

Thanks for contributing!

PR #219, opened 2019-02-04
Update re: plugin API changes

dralley:
* Use specialized serializer instead of ContentSerializer
* Fix post stages refactoring

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@
 from setuptools import setup, find_packages
 
 requirements = [
-    'pulpcore-plugin==0.1.0b16',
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -3,7 +3,7 @@
 from setuptools import setup, find_packages
 
 requirements = [
-    'pulpcore-plugin==0.1.0b16',
+    'pulpcore-plugin>=0.1.0b18',
```
asmacdo: Are we sure we don't want to pin?

asmacdo: After discussion on IRC, we decided to pin releases to pypi, but leave `master` as >=

PR #221, opened 2019-02-13
Fix the build of `content_host` URL using the distribution `base_path`

rochacbruno:
Fix the build of `content_host` URL using the distribution `base_path`

Required PR: https://github.com/PulpQE/pulp-smash/pull/1169

[noissue]


rochacbruno:
@pulp/qe

PR #225, opened 2019-03-08
Install pulpcore with postgres extra pacakges

daviddavis:
ref #4270


asmacdo:
These tests won't pass, but at least the tests are running now. Previously, the installation was failing.

PR #226, opened 2019-03-25
Use base serializer to create ContentArtifacts

asmacdo:
SingleArtifactSerializer was updated to handle ContentArtifacts so each
plugin doesn't need to.

Hashtag: @CodeHeeler ty 100 times

https://pulp.plan.io/issues/4451
fixes #4451


asmacdo:
This should fix 1 failing test, but not all of them. 

Failing:

    pulp_python/tests/functional/api/test_crud_content_unit.py::ContentUnitTestCase::test_01_create_content_unit FAILED

PR #232, opened 2019-04-11
Remove the conditionals for rtd theme

asmacdo:
Seems to be causing a problem with rendering the restapi.html
[noissue]


asmacdo:
I merged this change to my own fork, and the restapi.html was viewable.
https://asmacdo-pulp-python.readthedocs.io/en/latest/restapi.html

PR #233, opened 2019-04-11
Link to the newly published API docs

asmacdo:
Adding these reference links should make the user experience more
smooth, particularly when navigating both pulp-python and pulpcore docs.
[noissue]

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -1,8 +1,14 @@
 User Setup
 ==========
 
-Install pulp-python
--------------------
+Ansible Installer (Recommended)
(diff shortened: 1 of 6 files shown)


docs/installation.rst
```diff
@@ -1,8 +1,14 @@
 User Setup
 ==========
 
-Install pulp-python
--------------------
+Ansible Installer (Recommended)
+-------------------------------
```
asmacdo: This could be a separate commit if anyone cares. Just fixed it when I noticed it was missing.

PR #235, opened 2019-04-12
Fixes relative_path so pip can install packages

CodeHeeler:
Adds trailing slash to relative path so packages are at expected location.
Also updates docs to include $CONTENT_ADDR as this now differs from
$BASE_ADDR for our content app.

fixes #4677
https://pulp.plan.io/issues/4677

### DIFF
--- a/pulp_python/app/tasks/publish.py
+++ b/pulp_python/app/tasks/publish.py
@@ -83,7 +83,7 @@ def write_simple_api(publication):
         publication (pulpcore.plugin.models.Publication): A publication to generate metadata for
 
     """
-    simple_dir = 'simple'
+    simple_dir = 'simple/'
     os.mkdir(simple_dir)
     project_names = (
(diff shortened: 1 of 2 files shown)


pulp_python/app/tasks/publish.py
```diff
@@ -104,7 +104,7 @@ def write_simple_api(publication):
         index.write(template.render(context))
 
     index_metadata = models.PublishedMetadata(
-        relative_path=simple_dir,
+        relative_path=simple_dir + "/",
```
asmacdo: Why not set `simple_dir = 'simple/'` instead of adding the trailing slash twice?

asmacdo:
Nitpickiest nitpick. Don't bother updating for this one, but next time...

The subject line of the commit message should be in the imperative form. s/Fixes/Fix/. Mostly trivial, but it makes sense if you read https://chris.beams.io/posts/git-commit/

PR #236, opened 2019-04-15
Bump version to beta5, add release notes

CodeHeeler:
[noissue]


docs/release-notes/3.0.z.rst
```diff
@@ -8,16 +8,4 @@ might be made until Beta is over.
 3.0.0b4
 =======
 
-* Adds support for `pulpcore 3.0.0.rc1 <https://docs.pulpproject.org/en/3.0/nightly/release-notes/pulpcore/3.0.x.html#rc1>`_.
```
asmacdo: I think we need to keep the old release notes and include this fix in the release notes for 3.0.0b5

PR #242, opened 2019-04-26
Swith to Master/Detail Publications

asmacdo:
https://pulp.plan.io/issues/4699
fixes #4699


docs/_scripts/base.sh
```diff
@@ -0,0 +1,13 @@
+export BASE_ADDR=http://localhost:24817
+export CONTENT_ADDR=http://localhost:24816
+
+wait_for_pulp() {
+  unset CREATED_RESOURCE
+  local task_url=$1
+  while [ -z "$CREATED_RESOURCE" ]
```
asmacdo: This seems to fail sometimes, CREATED_RESOURCE gets set to `null`.

docs/_scripts/quickstart.sh
```diff
@@ -0,0 +1,19 @@
+# This script will execute the component scripts and ensure that the documented examples
+# work as expected.
```
asmacdo: Idea for another day:

It might be worth running quickstart.sh when building the docs, to test it out. As a possible side benefit, we will populate the db, which might allow us to include examples in the REST API docs.

PR #243, opened 2019-05-13
Add detail python distribution

asmacdo:
Master/Detail distributions are added to pulpcore in rc2, this patch is
necessary to keep up to date with pulpcore.

https://pulp.plan.io/issues/4797
fixes #4797


pulp_python/app/serializers.py
```diff
@@ -334,6 +345,13 @@ class PythonPublicationSerializer(core_serializers.PublicationSerializer):
     A Serializer for PythonPublication.
     """
 
+    python_distributions = core_serializers.DetailRelatedField(
```
gmbnomis: I think we want to establish a convention and call this field `distributions` in all plugins (@bmbouter?) (where it is applicable, i.e. there is a Publication).  For example, see the file plugin: https://github.com/pulp/pulp_file/blob/88f5bd302d1c98b7b34b474fe9bf8d6d9e7cce0c/pulp_file/app/serializers.py#L84

pulp_python/app/viewsets.py
```diff
@@ -40,6 +40,20 @@
 }
 
 
+class PythonDistributionViewSet(platform.BaseDistributionViewSet):
+    """
+    <!-- User-facing documentation, rendered as html-->
+    Pulp Python Distributions are used to distribute <a href="../restapi.html#tag/remotes">Python
+    Publications.</a> <b> Pulp Python Distributions should not be confused "Python Distribution" as
```
CodeHeeler: "should not be confused with"

PR #244, opened 2019-05-20
Bump version to beta6, add release notes

CodeHeeler:
[noissue]

### DIFF
--- a/docs/release-notes/3.0.z.rst
+++ b/docs/release-notes/3.0.z.rst
@@ -5,6 +5,24 @@ pulp-python 3.0 Release Notes
 pulp-python 3.0 is currently in Beta. Backwards incompatible changes
 might be made until Beta is over.
 
+3.0.0b6
+=======
+
+* See all changes `here <https://github.com/pulp/pulp_python/compare/3.0.0b5...3.0.0b6>`_.
+
+* Adds support for `pulpcore 3.0.0.rc2 <https://docs.pulpproject.org/en/3.0/rc/release-notes/pulpcore/3.0.x.html#rc2>`_.
+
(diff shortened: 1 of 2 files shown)


docs/release-notes/3.0.z.rst
```diff
@@ -5,6 +5,16 @@ pulp-python 3.0 Release Notes
...
 
+3.0.0b6
+=======
+
+* Adds support for `pulpcore 3.0.0.rc2 <https://docs.pulpproject.org/en/3.0/rc/release-notes/pulpcore/3.0.x.html#rc2>`_.
+
+3.0.0b5
+=======
+
+* Fix relative_path to allow pip install
```
asmacdo: Could you add a link to a redmine issue for this one?

asmacdo: Also, I think we should start including links like these: https://github.com/pulp/pulp_python/compare/3.0.0b4...3.0.0b5

CodeHeeler: That's not new, it's only showing up that way because I reordered our release notes to show the most recent first (b6->b5->b4) to match the way pulpcore does them: https://docs.pulpproject.org/en/3.0/rc/release-notes/pulpcore/3.0.x.html#rc2 instead of the way we had them before (b4->b5)

CodeHeeler: As for the release comparison links, I like that idea--is github reliable in it being that pattern since I can't otherwise create that link ahead of actually releasing b6?

asmacdo: Yeah, you can go ahead and say b6 and the link will work once its tagged.

PR #252, opened 2019-08-19
Update the travis config with the plugin_template.

asmacdo:
[noissue]

### DIFF
--- a/.travis/before_install.sh
+++ b/.travis/before_install.sh
@@ -15,7 +15,7 @@ export POST_BEFORE_INSTALL=$TRAVIS_BUILD_DIR/.travis/post_before_install.sh
 COMMIT_MSG=$(git log --format=%B --no-merges -1)
 export COMMIT_MSG
 
-if [ -x $PRE_BEFORE_INSTALL ]; then
+if [ -f $PRE_BEFORE_INSTALL ]; then
     $PRE_BEFORE_INSTALL
 fi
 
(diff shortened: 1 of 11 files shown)


.travis/before_install.sh
```diff
@@ -15,7 +15,7 @@ export POST_BEFORE_INSTALL=$TRAVIS_BUILD_DIR/.travis/post_before_install.sh
 COMMIT_MSG=$(git log --format=%B --no-merges -1)
 export COMMIT_MSG
 
-if [ -x $PRE_BEFORE_INSTALL ]; then
+if [ -f $PRE_BEFORE_INSTALL ]; then
```
dralley: What is the reasoning for this?

.travis/playbook.yml
```diff
@@ -28,7 +29,6 @@
       secret_key: 'secret'
       databases:
         default:
-          ENGINE: "{{ pulp_db_backend }}"
```
dralley: Where does the definition of ENGINE come from then, a defaults.yaml or something?

.travis/script.sh
```diff
@@ -66,6 +67,30 @@ if [ "$TEST" = 'bindings' ]; then
   pip install ./pulp_python-client
 
   python $TRAVIS_BUILD_DIR/.travis/test_bindings.py
+
```
dralley: I can't review this part, probably check with Dennis

.travis/playbook.yml
asmacdo: Yup: https://github.com/pulp/ansible-pulp/blob/master/roles/pulp-database/defaults/main.yml#L9

Though, since we no longer allow maria, its probably never going to be changed.

PR #261, opened 2019-10-10
Use JSONField

dralley:
[noissue]


dralley:
There's a transient test failure that is quite frustrating, I will file an issue

PR #265, opened 2020-01-20
skip project and project_metadata dir creation if dir exists

nbeernink:
This solves the problem in https://pulp.plan.io/issues/5098


nbeernink:
I reviewed the contributing guidelines but could not find a changes dir. Hope I'm doing this correctly, if not please let me know :)

PR #270, opened 2020-05-17
Removes Classifiers model and serializer

dkliban:
The classifiers are simply a list of strings. This patch converts this field into a JSONField.

fixes: #5452
https://pulp.plan.io/issues/5452


pulp_python/tests/functional/api/test_sync.py
```diff
@@ -87,9 +87,7 @@ def test_sync(self):
         sync(self.cfg, remote, repo)
         repo = self.client.get(repo['pulp_href'])
 
-        self.assertNotEqual(latest_version_href, repo['latest_version_href'])
-        self.assertDictEqual(get_content_summary(repo), PYTHON_XS_FIXTURE_SUMMARY)
-        self.assertDictEqual(get_added_content_summary(repo), {})
```
dralley, line 92: Why remove this?

dkliban: Because this is no longer true. Since a new 'no-op' version is not getting created, the 'added_content_summary' is no longer empty. The assertion about the 'content_summary' is already made on line 82.

dralley: Makes sense.

PR #289, opened 2020-06-06
updated sync.sh, publication.sh and distribution.sh scripts

eorochena:
Updated sync.sh, publication.sh and distribution.sh in docs/_scripts to reference wait_until_task_finished function from base.sh


eorochena:
closes 6918
https://pulp.plan.io/issues/6918

PR #291, opened 2020-06-06
updated sync.sh, publication.sh and distribution.sh scripts

eorochena:
[noissue]

### DIFF
--- a/CHANGES/6918.doc
+++ b/CHANGES/6918.doc
@@ -1,0 +1,1 @@
+Updated sync.sh, publication.sh and distribution.sh in docs/_scripts to reference wait_until_task_finished function from base.sh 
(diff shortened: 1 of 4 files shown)


eorochena:
This pull request is to address https://pulp.plan.io/issues/6918 but for some reason travis keeps failing even with the [noissue] label I will try to do the suggested changes 

with git add, commit --amend and push

Thanks

dralley:
The reason its unhappy is that there isn't a changelog entry, create a file called ```CHANGES/6918.doc``` and write your short description there. Then Travis should pass.

dralley:
Also, the reason adding ```[noissue]``` didn't work is because it's not in the commit message, just in the github description, and it needs to be in the commit message to be found in the automation.  But since there's an issue this is resolving, the appropriate thing is to add a changelog entry.

eorochena:
@dralley Thanks for the advise, all checks now passed.

PR #332, opened 2021-03-04
Update require_pulp_plugins

fao89:
[noissue]


fao89:
@pulp/python the test were being skipped, it enables the tests again, but I don't have the background for fixing these tests

dralley (approved):

fao89:
I rebased my PR, so I got this change:
https://github.com/pulp/pulp_python/pull/334/files
which uses http://pulp-fixtures:8080/
https://github.com/pulp/pulp_python/blob/master/.ci/ansible/smash-config.json#L34
but now the error is due https
what is the right fixture?

gerrod3:
> what is the right fixture?

Bandersnatch requires the remote url to be https, so the fixtures url must point to our hosted fixtures and not use the container built fixtures.

fao89:
> > what is the right fixture?
> 
> Bandersnatch requires the remote url to be https, so the fixtures url must point to our hosted fixtures and not use the container built fixtures.

if we set `docker-fixture: false` we will get `https://fixtures.pulpproject.org` but it may break CLI test:
https://github.com/pulp/plugin_template/commit/24fbd58ab19ea8f0ca6cb6173f55ac60c0d9ae6d
when you enable it:
https://github.com/pulp/pulp_python/blob/master/template_config.yml#L37

gerrod3:
Hmm, where is that variable docker_fixtures being set? We should turn that off so that it chooses the hosted fixtures instead.

template_config.yml
```diff
@@ -13,7 +13,7 @@ deploy_client_to_rubygems: true
 deploy_daily_client_to_pypi: true
 deploy_daily_client_to_rubygems: true
 deploy_to_pypi: true
-docker_fixtures: true
+docker_fixtures: false
```
fao89, line 17: I turned it off and ran plugin-template

fao89:
Finally green \o/
@pulp/python

PR #390, opened 2021-06-16
Remove api.json

goosemania:
It is no longer used for API docs after moving to pulprpoject.org.
Add to .gitignore as well.
Ensure that _static dir is created when needed.
Remove a hack for keeping an empty directory in the git repo.

[noissue]


goosemania:
It can be merged after docs moved to pulpproject.org.
This PR is a part of api.json cleanup https://pulp.plan.io/issues/8866

dralley (approved):

PR #416, opened 2021-08-20
Update CI + add upgrade tests

### LINKED ISSUE #408: Add upgrade tests
as developer I have tests assuring migrations work fine

fao89:
closes #408


template_config.yml
```diff
@@ -48,5 +60,7 @@ test_s3: false
 travis_addtl_services: []
 travis_notifications: None
 update_redmine: false
-upgrade_range: []
+upgrade_range:
+- pulp_python_branch: 3.2
```
fao89, line 64: @gerrod3 @dralley could you please create this branch?

gerrod3: Branch should be created now. Do you also need a 3.3 branch?

fao89: no, just the 3.2, I'll be adding the pre_upgrade tests there

gerrod3:
@fao89 You might want to remake this PR with the latest changes on master.

.github/workflows/scripts/script.sh
```diff
@@ -99,6 +99,56 @@ cmd_prefix bash -c "PULP_DATABASES__default__USER=postgres django-admin test --n
 export PYTHONPATH=$REPO_ROOT:$REPO_ROOT/../pulpcore${PYTHONPATH:+:${PYTHONPATH}}
 
 
+if [[ "$TEST" == "upgrade" ]]; then
+  # Handle app label change:
+  sed -i "/require_pulp_plugins(/d" pulp_python/tests/functional/utils.py
+
+  # Running pre upgrade tests:
+  pytest -v -r sx --color=yes --pyargs -capture=no pulp_python.tests.upgrade.pre
```
fao89, line 107: @gerrod3 here is where the upgrade tests will start to run

gerrod3:
@fao89 It seems that your upgrade tests don't account for changes in test_requirements.txt. Going from 3.2 -> 3.4 I think I added 1 or 2 requirements to run the tests.

pulp_python/tests/upgrade/post/test_publish.py
```diff
@@ -0,0 +1,131 @@
...
+        self.assertEqual(publication.repository_version, non_latest)
+
+        # Step 7
+        with self.assertRaises(ApiException):
+            body = {"repository": self.repo.pulp_href, "repository_version": non_latest}
+            self.publications.create(body)
+
+        # Step 8
+        self.assertIn("/pypi/pulp_post_upgrade_test/", distribution.base_url)
+
```
gerrod3, line 125: If these tests only run in the CI then the base url should be `https://pulp:80/pypi/pulp_post_upgrade_test/` I think. It's fine this way too. 

Also, I think the publication & distribution from the pre-test aren't destroyed correct? If not then should a test be added here to ensure you can access them from the new endpoint location?

pulp_python/app/utils.py
```diff
@@ -150,6 +150,23 @@ def latest_content_version(content_query, version):
     return latest_content
 
 
+def json_to_dict(data):
+    """
+    Converts a JSON string into a Python dictionary.
```
gerrod3, line 155: This is interesting. Why is it needed?

gerrod3 (commented):
Looks good, just a couple questions before I approve.

pulp_python/app/utils.py
fao89: the post-test was breaking because the data was already a dict:
https://github.com/pulp/pulp_python/runs/3406807006?check_suite_focus=true#step:12:645

pulp_python/tests/upgrade/post/test_publish.py
fao89: I think I can use the new endpoint when syncing, I'll change that

gerrod3 (approved):
Thanks!

PR #454, opened 2021-10-13
RFC: Adding 101 demo

fao89:
Ref: https://discourse.pulpproject.org/t/rfc-pulp-101-webpage/155/8

### DIFF
--- a/.github/workflows/demo.yml
+++ b/.github/workflows/demo.yml
@@ -1,0 +1,123 @@
+name: Pulp Python 101
+
+on:
+  schedule:
+    - cron: '30 2 * * 0'
+
+env:
+  COLORTERM: 'yes'
+  PY_COLORS: '1'
+  TERM: 'xterm-256color'
+
+jobs:
+  demo:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-python@v2
+        with:
+          python-version: "3.8"
+
+      - name: Starting Pulp server
+        run: |
+            mkdir settings pulp_storage pgsql containers
+            echo "CONTENT_ORIGIN='https://pulp.example.com'" >> settings/settings.py
+            docker run --detach \
+                      --name pulp \
+                      --publish 443:443 \
+                      --volume "/$(pwd)/settings:/etc/pulp:Z" \
+                      --volume "/$(pwd)/pulp_storage":/var/lib/pulp:Z \
+                      --volume "/$(pwd)/pgsql":/var/lib/pgsql:Z \
+                      --volume "/$(pwd)/containers":/var/lib/containers:Z \
+                      --device /dev/fuse \
+                      pulp/pulp:https
+
+            # Wait server to be up
+            sleep 10
+            for _ in $(seq 20)
+            do
+              sleep 2
+              if curl --insecure --fail https://localhost/pulp/api/v3/status/ > /dev/null 2>&1
+              then
+                break
+              fi
+            done
+
+            # Edit /etc/hosts
+            echo 127.0.0.1 pulp.example.com | sudo tee -a /etc/hosts
+
+            # Hack: adding pulp CA to certifi.where()
+            sudo docker cp pulp:/etc/pulp/certs/pulp_webserver.crt /usr/local/share/ca-certificates/pulp_webserver.crt
+            pip install certifi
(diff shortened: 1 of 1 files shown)


fao89:
@melcorr

.github/workflows/demo.yml
```diff
@@ -0,0 +1,123 @@
...
+                      --volume "/$(pwd)/containers":/var/lib/containers:Z \
+                      --device /dev/fuse \
+                      pulp/pulp:https
+
+            # Wait server to be up
+            sleep 10
+            for _ in $(seq 20)
+            do
+              sleep 2
+              if curl --insecure --fail https://localhost/pulp/api/v3/status/ > /dev/null 2>&1
```
fao89, line 39: This is the step where we set the Pulp server.
The other steps are client steps

.github/workflows/demo.yml
```diff
@@ -0,0 +1,123 @@
...
+            sudo update-ca-certificates
+
+            # Setting the password for admin
+            docker exec pulp bash -c 'pulpcore-manager reset-admin-password --password ci_password'
+
+      - name: Setting Pulp CLI
+        run: |
+          pip install pulp-cli[pygments]
+          pulp config create --username admin --base-url https://pulp.example.com --password ci_password
+          pulp status
```
fao89, line 69: Starting from here, these are the client steps.

Let's split the client steps into 3 types of users:
- admin
- regular user
- QE

Here we are setting the CLI for the admin, the person who will create remotes/repositories/publications/distributions/...

.github/workflows/demo.yml
```diff
@@ -0,0 +1,123 @@
...
+          echo "PUBLICATION_HREF=$PUBLICATION_HREF" >> $GITHUB_ENV
+
+      - name: Distribute "dev" packages
+        run: |
+          pulp python distribution create --name "dev" --base-path "dev" --publication $PUBLICATION_HREF
+
+      - name: Publish & Distribute "prod" packages
+        run: |
+          PROD_PUBLICATION_HREF=$(pulp python publication create --repository "prod" | jq -r .pulp_href)
+          pulp python distribution create --name "prod" --base-path "prod" --publication $PROD_PUBLICATION_HREF
```
fao89, line 101: Client-side - where admin user is prepping pulp to be used by QE/regular users

.github/workflows/demo.yml
```diff
@@ -0,0 +1,123 @@
...
+
+      - name: Install package from pulp "dev" repository
+        run: |
+          pip install -vvv --trusted-host pulp -i https://pulp.example.com/pypi/dev/simple/ shelf-reader
```
fao89, line 105: Client-side, QE user testing dev content

(consider pip installing from dev distribution as a test)

Here I could explore more and add some tests

.github/workflows/demo.yml
```diff
@@ -0,0 +1,123 @@
...
+
+      - name: Sync packages from "dev" to "prod"
+        run: |
+          pulp python repository sync --name "prod" --remote "prod"
```
fao89, line 109: Client-side, QE user

Here is the part where the content is "promoted" from dev to prod repository

Maybe I should start another Pulp CLI with QE user?

.github/workflows/demo.yml
```diff
@@ -0,0 +1,123 @@
...
+
+      - name: Install package from pulp "prod" repository
+        run: |
+          pip install -vvv --force-reinstall --trusted-host pulp -i https://pulp.example.com/pypi/prod/simple/ shelf-reader
```
fao89, line 113: Client-side, regular user

User pip installing python packages from pulp prod distribution

.github/workflows/demo.yml
melcorr: Is it realistic that this is part of the same GHA script?

.github/workflows/demo.yml
melcorr: Yeah that is what I think we mean when we say CI/CD scenario. 
Integration tests when a dev pushes changes at this point. 
Sorry for making work for you @fao89  :(

fao89: this is a PoC, we are learning the best way to communicate with users. We'll probably have a lot more changes!

PR #637, opened 2023-12-06
Set SERVE_FROM_PUBLICATION on PythonDistribution

daviddavis:
I don't know if this is strictly required since pulp_python defines its own content handler function but SERVE_FROM_PUBLICATION ought to be set from a technical correctness standpoint. Also future work might expect it to be set such as the repo version protection PR I have open against pulpcore:

https://github.com/pulp/pulpcore/pull/4750

### DIFF
--- a/pulp_python/app/models.py
+++ b/pulp_python/app/models.py
@@ -52,6 +52,7 @@ class PythonDistribution(Distribution):
     """
 
     TYPE = "python"
+    SERVE_FROM_PUBLICATION = True
 
     allow_uploads = models.BooleanField(default=True)
 


daviddavis:
More info about SERVE_FROM_PUBLICATION: https://github.com/pulp/pulpcore/commit/27a0dc7424769b2d5d7e98ebc4cf9cdba6cdc149

gerrod3:
I'm surprised this didn't break any tests. The reason the python distribution is not marked `SERVE_FROM_PUBLICATION` is because you don't need a publication to serve python repositories. Back in 3.5 I introduced the PyPI apis with the new setting `PYPI_API_HOSTNAME` to replace the publication based distribution. The new apis would figure out what the client needs and redirect to the content app for the final file. However to not break those that were using publications I had the new apis automatically redirect to the content app when using publications: https://github.com/pulp/pulp_python/blob/main/pulp_python/app/pypi/views.py#L87. Thus to support both use cases of publications and no-publications I didn't mark the distribution as such as doing so **_should_** stop the content app from serving the files when no publication is used. I think this shows that we do not have a test for the no-publication use case (or that something is broken in pulpcore?, unlikely since the change you link has a test for that).

daviddavis:
@gerrod3 that's interesting. I didn't realize that it was possible to use pulp_python without publications. so users can set a repo on a distribution and that'll distribute the repo's packages without having to publish the repo? if that's the case I can close out this PR. 

The one concern I'd have is that without setting this though [the repo version protection feature](https://github.com/pulp/pulpcore/pull/4750) may not work properly for pulp_python repos. The code will save the repo's latest version instead of the latest *published* version. Is that the correct behavior? If not, we could override the `protected_versions` method for `PythonRepository`.

PR #654, opened 2024-03-24
Update CI files for branch main

pulpbot:


ggainey:
pulpcore lowerbound is 3.25, which is not a supported branch. We can either move that LB to a supported branch that has the tablib pin, or we can pin tablib explicitly here and release.

gerrod3 (approved):

PR #683, opened 2024-06-25
Update pkginfo library to support newer python packages

### LINKED ISSUE #682: Support Metadata 2.3 | update version of pkginfo
Several Python packages, such as `urllib3-2.2.2-py3-none-any.whl`, use metadata version 2.3. We cannot upload these packages to Pulp using the following command:

```bash 
pulp content upload --file "urllib3-2.2.2-py3-none-any.whl" --relative-path "urllib3-2.2.2-py3-none-any.whl"
```

This attempt fails with the error:

```bash
failed: 'null value in column "version" of relation "python_pythonpackagecontent" violates not-null constraint
DETAIL: Failing row contains (01904e82-d1c0-7fa8-9fff-1df6899fceef, urllib3-2.2.2-py3-none-any.whl, bdist_wheel, , null, 2.3, , <h1 align="center"
```

After repacking `urllib3-2.2.2-py3-none-any.whl` with `Metadata-Version: 2.1` or `Metadata-Version: 2.2`, the content uploads successfully.

The issue appears to be with the `pkginfo` library, which does not support metadata 2.3 (this support is only available from version 1.10.0 onwards)."

dralley:
closes #682


dralley:
Might need some fields added as well

dralley:
I need to find a package that uses metadata version 2.3 and also uses this dynamic feature, so that we can properly test it.

dralley:
Looks like the `twine` package happens to have `dynamic` set.

dralley:
I don't mean dynamic fixtures necessarily, even a pre-generated package in the fixtures would be fine

But I assume the answer there is also probably no.  That's fine, I can make do

pulp_python/app/models.py
```diff
@@ -178,6 +182,7 @@ class PythonPackageContent(Content):
     classifiers = models.JSONField(default=list)
     project_urls = models.JSONField(default=dict)
     description_content_type = models.TextField()
+    dynamic = models.TextField()
```
gerrod3, line 185: I think we are missing `provides_extra`. Added in 2.1. Also, can we reformat the list to be like pkginfo? https://github.com/jwilk-mirrors/python-pkginfo/blob/master/pkginfo/distribution.py#L124

dralley: Sure. For `provides_extra`, are we going to be missing metadata for existing packages?  I don't think we ever had a problem with rejection of metadata 2.1 packages as we did with 2.3

How do you want to deal with that?

dralley: Also, I filed https://github.com/pypi/warehouse/issues/16190 because I'm having difficulty figuring out exactly what these fields are meant to look like.  The core metadata specification describes how they're represented in the metadata files within a package, not how they're represented on PyPI.

gerrod3: > Sure. For `provides_extra`, are we going to be missing metadata for existing packages? I don't think we ever had a problem with rejection of metadata 2.1 packages as we did with 2.3
> 
> How do you want to deal with that?

We might need to create a new endpoint that runs a task to fix bad metadata. Unless you are uploading every package most python content will have incorrect metadata because syncing doesn't individually check the metadata for each package. Currently we take the latest's package metadata and re-use it for every version of that package.

PR #751, opened 2024-10-24
Fix Any serializer type regression breaking Ruby bindings

pedro-psb:
See context: https://github.com/pulp/pulp_rpm/issues/3639


pedro-psb:
Does this needs the `backport`  label too?

gerrod3 (approved):

PR #814, opened 2025-03-05
Pin bandersnatch to <6.6.

ggainey:
See https://github.com/pypa/bandersnatch/issues/1892 for the discussion.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ requires-python = ">=3.9"
 dependencies = [
   "pulpcore>=3.49.0,<3.85",
   "pkginfo>=1.10.0,<1.13.0",
-  "bandersnatch>=6.3,<7.0",  # Anything >6.3 requires Python 3.10+
+  "bandersnatch>=6.3.0,<6.4",  # Anything >6.3 requires Python 3.10+
   "pypi-simple>=1.5.0,<2.0",
 ]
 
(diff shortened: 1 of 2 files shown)


ggainey:
The install-failure here is due to our more-strict upperbounds check requiring bandersnatch 6.5, on an image built with python-3.9 - which is *not supported* by bandersnatch>=6.4.0. We will need to pin to ~=6.3.0 until we stop supporting py3.9, and/or figure out how to make the UB check take the python-version into account.

pyproject.toml
```diff
@@ -29,7 +29,7 @@ requires-python = ">=3.9"
 dependencies = [
   "pulpcore>=3.49.0,<3.85",
   "pkginfo>=1.10.0,<1.13.0",
-  "bandersnatch>=6.3,<7.0",  # Anything >6.3 requires Python 3.10+
+  "bandersnatch~=6.3.0",  # Anything >6.3 requires Python 3.10+
```
mdellweg, line 32: Should this be
```suggestion
  "bandersnatch>=6.3.0,<6.4",  # Anything >6.3 requires Python 3.10+
```
then?

ggainey: Aren't these identical?

mdellweg: Yes, and no.
They mean the same to python, but dependabot treats them differently.

mdellweg (approved):

PR #840, opened 2025-04-24
Fix formatting in changelog

jobselko:
This PR fixes broken version headers in the changelog and removes redundant empty lines.


mdellweg:
Thank you!

Note to us: If we see the collect script confused and wanting to add these versions again, we need to fix the changelog on those branches...

PR #841, opened 2025-04-28
Add repair_metadata endpoint to Repositories

### LINKED ISSUE #805: Add repair endpoint to repositories
With the new repair metadata script we should also have a repository endpoint that allows for repairing the metadata of a repository's content through a task. Currently the repair script only handles immediate content, but we could also repair on-demand content if the associated remote supported the PyPI json version API.

jobselko:
This PR adds a new `repair_metadata` endpoint to Repositories, which repairs a repository's content metadata through a task. Currently, it supports only immediate content, on-demand support will be added later.

Closes #805

### DIFF
--- a/pulp_python/app/tasks/repair.py
+++ b/pulp_python/app/tasks/repair.py
@@ -1,0 +1,89 @@
+import logging
+import uuid
+from gettext import gettext as _
+
+from django.db.models.query import QuerySet
+from pulpcore.plugin.models import ProgressReport
+from pulpcore.plugin.util import get_domain
+
+from pulp_python.app.models import PythonPackageContent, PythonRepository
+from pulp_python.app.utils import artifact_to_python_content_data
+
+log = logging.getLogger(__name__)
+
+
+def repair(repository_pk: uuid.UUID) -> None:
+    """
+    Repairs metadata of all packages for the specified repository.
+
+    Args:
+        repository_pk (uuid.UUID): The primary key of the repository to repair.
+
+    Returns:
+        None
+    """
+    repository = PythonRepository.objects.get(pk=repository_pk)
+
+    log.info(
+        _(
+            "Repairing packages' metadata for the latest version of repository {}."
+        ).format(repository.name)
+    )
+    content_set = repository.latest_version().content.values_list("pk", flat=True)
+    content = PythonPackageContent.objects.filter(pk__in=content_set)
+
+    num_repaired = repair_metadata(content)
+    log.info(_("{} packages' metadata repaired.").format(num_repaired))
+
+
+def repair_metadata(content: QuerySet[PythonPackageContent]) -> int:
+    """
+    Repairs metadata for a queryset of PythonPackageContent objects
+    and updates the progress report.
+
+    Args:
+        content (QuerySet[PythonPackageContent]): The queryset of items to repair.
+
+    Returns:
+        int: The number of packages that were repaired.
+    """
+    # TODO: Add on_demand content repair
+    immediate_content = content.filter(contentartifact__artifact__isnull=False)
+    domain = get_domain()
+
+    batch = []
+    set_of_update_fields = set()
+    total_repaired = 0
+
+    progress_report = ProgressReport(
+        message="Repairing packages' metadata",
+        code="repair.metadata",
+        total=immediate_content.count(),
+    )
+    progress_report.save()
+    with progress_report:
+        for package in progress_report.iter(
+            immediate_content.prefetch_related("_artifacts").iterator(chunk_size=1000)
+        ):
+            new_data = artifact_to_python_content_data(
+                package.filename, package._artifacts.get(), domain
+            )
+            changed = False
+            for field, value in new_data.items():
+                if getattr(package, field) != value:
+                    setattr(package, field, value)
+                    set_of_update_fields.add(field)
+                    changed = True
+            if changed:
+                batch.append(package)
+            if len(batch) == 1000:
+                total_repaired += len(batch)
+                PythonPackageContent.objects.bulk_update(batch, set_of_update_fields)
+                batch = []
+                set_of_update_fields.clear()
+
+    if batch:
+        total_repaired += len(batch)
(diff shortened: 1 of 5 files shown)


pulp_python/app/tasks/repair.py
```diff
@@ -0,0 +1,83 @@
...
+    content = PythonPackageContent.objects.filter(pk__in=content_set)
+    num_repaired = repair_metadata(content)
+    log.info(
+        _(
+            f"{len(content_set)} packages processed, {num_repaired} package metadata repaired."
+        )
+    )
+
+
+def repair_metadata(content: ContentQuerySet) -> int:
```
jobselko, line 40: I will probably need to remove the `ContentQuerySet` type hint, since it is not imported in `pulpcore/pulpcore/app/models/__init__.py`. Would it make sense to add it to pulpcore and wait for the next release instead?

pulp_python/app/tasks/repair.py
```diff
@@ -0,0 +1,85 @@
...
+        ).format(repository.name)
+    )
+    content_set = repository.latest_version().content.values_list("pk", flat=True)
+    content = PythonPackageContent.objects.filter(pk__in=content_set)
+    num_repaired = repair_metadata(content)
+    log.info(
+        _("{} packages processed, {} package metadata repaired.").format(
+            len(content_set), num_repaired
+        )
+    )
```
gerrod3, line 38: Let's use ProgressReports (https://github.com/pulp/pulpcore/blob/main/pulpcore/app/models/progress.py#L19) to report on the number of packages repaired.

pulp_python/app/tasks/repair.py
```diff
@@ -0,0 +1,85 @@
...
+    Repairs metadata for a queryset of PythonPackageContent objects.
+
+    Args:
+        content (ContentQuerySet): The queryset of PythonPackageContent items to repair.
+
+    Returns:
+        int: The number of packages that were repaired.
+    """
+    # TODO: Add on_demand content repair
+    os.chdir(settings.WORKING_DIRECTORY)
```
gerrod3, line 53:
```suggestion
```
Tasks get a temporary directory to run in when dispatched.

pulp_python/app/tasks/repair.py
```diff
@@ -0,0 +1,85 @@
...
+    content = content.select_related("pulp_domain")
```
gerrod3, line 54:
```suggestion
```
Tasks run in one domain (the domain they are dispatched from), so you don't need to fetch each content's domain, they're all the same! You can use `get_domain()` to get it if needed.

pulp_python/app/viewsets.py
```diff
@@ -122,6 +122,26 @@ class PythonRepositoryViewSet(
...
     }
 
+    @extend_schema(
+        summary="Repair metadata",
+        responses={202: AsyncOperationResponseSerializer},
+    )
+    @action(detail=True, methods=["post"], serializer_class=None)
+    def repair_metadata(self, request, pk):
+        """
+        <!-- User-facing documentation, rendered as html-->
```
gerrod3, line 132:
```suggestion
```

gerrod3 (changes requested):

pulp_python/app/tasks/repair.py
jobselko: Solved by using more generic `QuerySet`, since I do not use any methods specific to `ContentQuerySet`

pulp_python/app/tasks/repair.py
```diff
@@ -0,0 +1,86 @@
...
+    content_set = repository.latest_version().content.values_list("pk", flat=True)
+    content = PythonPackageContent.objects.filter(pk__in=content_set)
+    num_repaired = repair_metadata(content)
+    progress_report = ProgressReport(
+        message=_("Repaired packages' metadata"),
+        code="repair.metadata",
+        state=TASK_STATES.COMPLETED,
+        total=len(content_set),
+        done=num_repaired,
+    )
```
gerrod3, line 41: I think this progress report should be updated as the task repairs content units. I would create this report before the repair and pass it onto the function so that the function can update the report as it goes.

pulp_python/app/tasks/repair.py
```diff
@@ -0,0 +1,86 @@
...
+
+    batch = []
+    set_of_update_fields = set()
+    total_repaired = 0
+
+    for package in immediate_content.prefetch_related("_artifacts").iterator(
+        chunk_size=1000
+    ):
+        new_data = artifact_to_python_content_data(
+            package.filename, package._artifacts.get(), package.pulp_domain
```
gerrod3, line 66:
```suggestion
            package.filename, package._artifacts.get(), domain
```
Where `domain = get_domain()`. All of the packages are a part of the same domain, so you should use the utility method to set this value, cause using `package.pulp_domain` would cause a n+1 query issue.

gerrod3 (changes requested):

pulp_python/app/tasks/repair.py
jobselko: I chose to update the progress report every 100 processed packages instead of one by one, since `bulk_update` happens every 1000 updated packages. But I am not sure if this is correct.

mdellweg: Have a look at the Progress reports implementation. When used with a context manager they do some sort of deferred update.

pulp_python/app/tasks/repair.py
```diff
@@ -0,0 +1,89 @@
...
+    domain = get_domain()
+
+    batch = []
+    set_of_update_fields = set()
+    total_repaired = 0
+
+    progress_report = ProgressReport(
+        message="Repairing packages' metadata",
+        code="repair.metadata",
+        total=len(immediate_content),
```
gerrod3, line 61: When you call `len` on the queryset you actually end up evaluating the whole queryset.
```suggestion
        total=immediate_content.count(),
```

gerrod3 (changes requested):
One last change and then this should be good to merge

gerrod3 (approved):
Thanks!

PR #956, opened 2025-09-04
[PULP-715] Add JSON-based Simple API

### LINKED ISSUE #625: Add support for new JSON simple index.
**Describe the solution you'd like**
Add support for the new JSON simple index https://peps.python.org/pep-0691/ for our publications + distributions.

Bandersnatch>=6.0.0 already supports writing the new format of simple pages. Look into using their existing code.

jobselko:
closes #625


pulp_python/tests/functional/api/test_pypi_simple_json_api.py
```diff
@@ -0,0 +1,71 @@
...
+    detail_url = f"{index_url}aiohttp"
+    headers = {"Accept": "application/vnd.pypi.simple.v1+json"}
+
+    response_index = requests.get(index_url, headers=headers)
+    data_index = response_index.json()
+    assert data_index["meta"] == {"api-version": "1.4", "_last-serial": 1000000000}
+    assert data_index["projects"]
+    for project in data_index["projects"]:
+        for i in ["_last-serial", "name"]:
+            assert project[i]
```
gerrod3, line 25:
```suggestion
            assert i in project
```

pulp_python/tests/functional/api/test_pypi_simple_json_api.py
```diff
@@ -0,0 +1,71 @@
...
+    assert response_index.headers["X-PyPI-Last-Serial"] == "1000000000"
+
+    response_detail = requests.get(detail_url, headers=headers)
+    data_detail = response_detail.json()
+    assert data_detail["meta"] == {"api-version": "1.4", "_last-serial": 1000000000}
+    assert data_detail["name"] == "aiohttp"
+    assert data_detail["files"]
+    for file in data_detail["files"]:
+        for i in ["filename", "url", "hashes"]:
+            assert file[i]
```
gerrod3, line 36:
```suggestion
            assert i in file
```

pulp_python/app/pypi/views.py
```diff
@@ -235,14 +252,55 @@ class SimpleView(PackageUploadMixin, ViewSet):
...
+            ]
+        else:
+            return [JSONRenderer(), BrowsableAPIRenderer()]
+
+    def handle_exception(self, exc):
+        # todo: fixes test_pull_through_filter?
+        if isinstance(exc, (Http404, NotFound)):
+            # Force JSON response since JSONRenderer is not in get_renderers()
+            return JsonResponse({"detail": str(exc)}, status=404)
+        return super().handle_exception(exc)
```
gerrod3, line 283: There's no reason the 404 has to be a JSON response. You could change the test to just check for the error string instead of expecting the result to be json.

pulp_python/app/pypi/views.py
```diff
@@ -235,14 +252,55 @@ class SimpleView(PackageUploadMixin, ViewSet):
...
+        """
+        if self.action in ["list", "retrieve"]:
+            # Ordered by priority if multiple content types are present
+            return [
+                TemplateHTMLRenderer(),
+                PyPISimpleHTMLRenderer(),
+                PyPISimpleJSONRenderer(),
+            ]
+        else:
+            return [JSONRenderer(), BrowsableAPIRenderer()]
```
gerrod3, line 276: Does `super().get_renders()` return these values?

pulp_python/app/utils.py
```diff
@@ -414,6 +418,46 @@ def write_simple_detail(project_name, project_packages, streamed=False):
...
+        "meta": {"api-version": PYPI_API_VERSION, "_last-serial": PYPI_SERIAL_CONSTANT},
+        "projects": projects,
+    }
+
+
+# todo: fields
+def write_simple_detail_json(project_name, project_packages):
+    """Writes the simple detail page in JSON format."""
+    files = []
+    for filename, url, sha256 in project_packages:
```
gerrod3, line 435: It might make sense to change project_packages to a list of dictionaries, or namedtuples, and then modify the simple_detail template accordingly.

gerrod3 (commented):
This PR looks like it is coming along well.

pulp_python/app/pypi/views.py
jobselko: Yes

pulp_python/app/utils.py
```diff
@@ -414,6 +426,53 @@ def write_simple_detail(project_name, project_packages, streamed=False):
...
+                    package["yanked_reason"]
+                    if package["yanked"] and package["yanked_reason"]
+                    else package["yanked"]
+                ),
+                # gpg-sig (not in warehouse)
+                # todo (from new PEPs):
+                # size (v1.1, PEP 700)
+                # upload-time (v1.1, PEP 700)
+                # core-metadata (PEP 7.14)
+                # provenance (v1.3, PEP 740)
```
gerrod3: Let's leave these for future PRs

pulp_python/app/tasks/publish.py
```diff
@@ -118,6 +118,7 @@ def write_project_page(name, simple_dir, package_releases, publication):
     metadata_relative_path = f"{project_dir}index.html"
 
     with open(metadata_relative_path, "w") as simple_metadata:
+        # todo?
```
gerrod3: Let's leave this for the future. We might not implement the feature for publications if we are planning to remove them for Pulp 4.

pulp_python/app/models.py
```diff
@@ -192,6 +192,9 @@ class PythonPackageContent(Content):
     packagetype = models.TextField(choices=PACKAGE_TYPES)
     python_version = models.TextField()
     sha256 = models.CharField(db_index=True, max_length=64)
+    sha256_metadata = models.CharField(max_length=64)
```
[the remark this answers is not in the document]
dralley: Should there be a management command capable of downloading the packages and updating the values from the packages?

pulp_python/app/utils.py
dralley: do you want the comments wholly removed?  I can create new issues for them if they don't already exist

pulp_python/app/pypi/views.py
```diff
@@ -252,7 +295,12 @@ def parse_package(release_package):
             stripped_url = urlunsplit(chain(parsed[:3], ("", "")))
             redirect_path = f"{path}/{release_package.filename}?redirect={stripped_url}"
             d_url = urljoin(self.base_content_url, redirect_path)
-            return release_package.filename, d_url, release_package.digests.get("sha256", "")
+            return {
+                "filename": release_package.filename,
+                "url": d_url,
+                "sha256": release_package.digests.get("sha256", ""),
+                # todo: more fields?
```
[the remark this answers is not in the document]
dralley: ack

pulp_python/app/models.py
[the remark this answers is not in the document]
gerrod3: Maybe, we have a repair command and endpoint that could do this.

pulp_python/app/utils.py
```diff
@@ -12,10 +12,13 @@
 from pulpcore.plugin.models import Remote
 
 
+# todo: why upper case?
 PYPI_LAST_SERIAL = "X-PYPI-LAST-SERIAL"
 """TODO This serial constant is temporary until Python repositories implements serials"""
 PYPI_SERIAL_CONSTANT = 1000000000
 
+PYPI_API_VERSION = "1.0"
```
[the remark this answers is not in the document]
gerrod3: I think the rename is better since this constant is for the simple api, both the html and json versions: https://packaging.python.org/en/latest/specifications/simple-repository-api/ and it falls under the /simple/ urls. 

There is a separate PyPI json api (that we use for syncing) that is "unique" to PyPI (https://docs.pypi.org/api/json/) and falls under the /pypi/ urls, which we also implement so we can perform Pulp-to-Pulp syncing.

If you don't like `SIMPLE_API` you can also name it `INDEX_API` since that is also what it is called.

pulp_python/app/tasks/publish.py
gerrod3: No, it's not deprecated in the sense that we will still fix issues with it, but we will point users to use the new preferred method if they want all the latest features.

pulp_python/app/models.py
```diff
@@ -192,6 +192,9 @@ class PythonPackageContent(Content):
...
+    yanked = models.BooleanField(default=False)
```
dralley: @gerrod3 So, I take it that the python plugin does not have "immutable" content, since "yanked" is explicitly mutable?

> The yanked attribute is not immutable once set, and may be rescinded in the future (and once rescinded, may be reset as well). Thus API users MUST be able to cope with a yanked file being “unyanked” (and even yanked again).

pulp_python/app/models.py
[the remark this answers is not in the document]
gerrod3: Also, what do you think of naming this `metadata_sha256`? Which do you think sounds better?

pulp_python/app/utils.py
```diff
@@ -38,8 +41,8 @@
 </head>
 <body>
     <h1>Links for {{ project_name }}</h1>
-    {% for name, path, sha256 in project_packages %}
-    <a href="{{ path }}#sha256={{ sha256 }}" rel="internal">{{ name }}</a><br/>
+    {% for pkg in project_packages %}
+    <a href="{{ pkg.url }}#sha256={{ pkg.sha256 }}" rel="internal">{{ pkg.filename }}</a><br/>
```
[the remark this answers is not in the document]
jobselko: I would prefer to make only the necessary changes to the HTML template for now.

pulp_python/app/utils.py
```diff
@@ -158,9 +160,9 @@ def parse_metadata(project, version, distribution):
     package["url"] = distribution.get("url") or ""
     package["sha256"] = distribution.get("digests", {}).get("sha256") or ""
     package["python_version"] = distribution.get("python_version") or package.get("python_version")
-    package["requires_python"] = distribution.get("requires_python") or package.get(
-        "requires_python"
-    )  # noqa: E501
+    package["requires_python"] = distribution.get("requires_python") or ""  # package.get(
+    #     "requires_python"
+    # )  # noqa: E501
```
jobselko: Keeping `package.get("requires_python")` returns `>=3.10` instead of `None` from `distribution.get("requires_python")`. Mixing project metadata and distribution data causes issue here.

See:
```
curl -s -H "Accept: application/vnd.pypi.simple.v1+json" https://pypi.org/simple/django/ | jq '.files[] | select(.filename == "Django-1.10.4-py2.py3-none-any.whl")'
...
"requires-python": null,
...
```
```
curl -s https://pypi.org/pypi/django/1.10.4/json | jq -r '.info.requires_python'

```
```
curl -s https://pypi.org/pypi/django/json | jq -r '.info.requires_python'
>=3.10
```
Should we always rewrite project metadata by distribution data?

pulp_python/tests/functional/api/test_pypi_simple_json_api.py
```diff
@@ -0,0 +1,126 @@
...
+    )
+    assert file_whl is not None, "wheel file not found"
+    assert file_whl["url"]
+    assert file_whl["hashes"] == {
+        "sha256": "2eceb1643c10c5e4a65970baf63bde43b79cbdac7de81dae853ce47ab05197e9"
+    }
+    assert file_whl["requires-python"] is None
+    assert file_whl["data-dist-info-metadata"] == {
+        "sha256": "ed333f0db05d77e933a157b7225b403ada9a2f93318d77b41b662eba78bac350"
+    }
```
jobselko: The value is here, and yet it is not ¯\_(ツ)_/¯

gerrod3 (approved):
Wow, great work! Thank you @jobselko :tada:

pulp_python/app/utils.py
```diff
@@ -187,6 +192,25 @@ def get_project_metadata_from_file(filename):
...
+        return ""
+    try:
+        with zipfile.ZipFile(filename, "r") as f:
+            for file_path in f.namelist():
+                if file_path.endswith(".dist-info/METADATA"):
+                    metadata_content = f.read(file_path)
+                    return hashlib.sha256(metadata_content).hexdigest()
+    except (zipfile.BadZipFile, KeyError, OSError):
+        pass
+    return ""
```
dralley: I wouldn't think we would ever want to return empty string?  Maybe "real checksum, else `None`"?  The field was made nullable, so empty string would be like a 2nd null value type.

pulp_python/app/utils.py
```diff
@@ -403,17 +427,65 @@ def find_artifact():
...
+        "name": canonicalize_name(project_name),
+        "files": [
+            {
+                # v1.0, PEP 691
+                "filename": package["filename"],
+                "url": package["url"],
+                "hashes": {"sha256": package["sha256"]},
+                "requires-python": package["requires_python"] or None,
+                # data-dist-info-metadata is deprecated alias for core-metadata
+                "data-dist-info-metadata": (
```
dralley: Does this fully take care of https://github.com/pulp/pulp_python/issues/997 ?

pulp_python/app/utils.py
```diff
@@ -403,17 +427,65 @@ def find_artifact():
...
+                    {"sha256": package["metadata_sha256"]} if package["metadata_sha256"] else False
+                ),
+                # yanked and yanked_reason are not implemented because they are mutable
+                # TODO in the future:
+                # size, upload-time (v1.1, PEP 700)
+                # core-metadata (PEP 7.14)
```
dralley: See above

pulp_python/app/utils.py
dralley: Oh you know what, brain fart, never mind.  #997 is completely unrelated.

pulp_python/app/utils.py
jobselko: Good catch! I also removed `package["metadata_sha256"] = package.get("metadata_sha256") or ""`, because the `package` already contains `metadata_sha256` - it comes from `package = parse_project_metadata(project)`

PR #981, opened 2025-09-29
experiment: django5

pedro-psb:
lets see what breaks.


pedro-psb:
Upper and lowerbound didn't install because of constraint checks. Other than than, azure and s3 fails one test, which is the same the nightly: https://github.com/pulp/pulp_python/actions/runs/18117595510/job/51556315665.

Although not 100% conclusive, evidence suggest we should be ok with django 5 here.

PR #983, opened 2025-10-01
Adds repository_version to the Python Distribution API.

dkliban:
This allows serving a specific repository version.


pedro-psb:
The `test_pull_through_with_repo` is failing on nightly too, probably unrelated.

gerrod3:
@dkliban I finally fixed the CI issue (it was multiple things biting us a once), but rebase and we should be good.

Issue #1012, opened 2025-11-12
Add vulnerability reports to python plugin


git-hyagi:
**Is your feature request related to a problem? Please describe.**
Support for the vulnerability reports along with documentation was added to the pulpcore in https://github.com/pulp/pulpcore/pull/6774 and the python plugin should build on it.

ref: https://issues.redhat.com/browse/PULP-823

PR #1013, opened 2025-11-12
Implements the vulnerability report

git-hyagi:
closes: #1012


pulp_python/app/viewsets.py
```diff
@@ -624,3 +651,27 @@ def create(self, request):
...
+
+
+def _build_osv_data(name, ecosystem, version=None):
+    """
+    Build an OSV data structure for vulnerability queries.
+    """
+    osv_data = {"package": {"name": name, "ecosystem": ecosystem}}
+    if version:
+        osv_data["version"] = version
+    return osv_data
```
gerrod3: Let's move this into a separate file under the `tasks/` directory.

pulp_python/app/viewsets.py
```diff
@@ -624,3 +651,27 @@ def create(self, request):
...
         )
         return core_viewsets.OperationPostponedResponse(result, request)
+
+
+async def get_repo_version_content(repo_version_pk: str):
+    """
+    Retrieve Python package content from a repository version for vulnerability scanning.
+    """
+    repo_version = await sync_to_async(RepositoryVersion.objects.get)(pk=repo_version_pk)
+    content_units = python_models.PythonPackageContent.objects.filter(pk__in=repo_version.content)
```
gerrod3:
```suggestion
    repo_version = await RepositoryVersion.objects.aget(pk=repo_version_pk)
    content_units = python_models.PythonPackageContent.objects.filter(pk__in=repo_version.content).values("name", "version")
```

git-hyagi: For the 
```
content_units = python_models.PythonPackageContent.objects.filter(pk__in=repo_version.content).values("name", "version")
```
I think this will not work because we also need the `content` objects found:
https://github.com/pulp/pulp_python/pull/1013/files#diff-38cc3b67ebd1f8ea44002fc7897552b1674435d85771f6ab27a0c86e383d00aeR666
https://github.com/pulp/pulpcore/blob/d9651981e20ac5b2dc1997f181204924f5c36cce/pulpcore/app/tasks/vulnerability_report.py#L133-L134

PR #1031, opened 2025-11-21
[PULP-946] Implement PEP 658

### LINKED ISSUE #1047: Implement PEP 658 – Serve Distribution Metadata in the Simple Repository API
[PEP 658](https://peps.python.org/pep-0658/)

> This PEP proposes adding an anchor tag to expose the METADATA file from distributions in the [PEP 503](https://peps.python.org/pep-0503/) “simple” repository API. A data-dist-info-metadata attribute is introduced to indicate that the file from a given distribution can be independently fetched.

jobselko:
closes #1047

### DIFF
--- a/pulp_python/app/migrations/0019_create_missing_metadata_artifacts.py
+++ b/pulp_python/app/migrations/0019_create_missing_metadata_artifacts.py
@@ -1,0 +1,201 @@
+# Generated manually on 2025-12-15 14:00 for creating missing metadata artifacts
+
+from django.db import migrations
+
+BATCH_SIZE = 1000
+
+
+def pulp_hashlib_new(name, *args, **kwargs):
+    """
+    Copied and updated (to comply with migrations) from pulpcore.
+    """
+    import hashlib as the_real_hashlib
+    from django.conf import settings
+
+    if name not in settings.ALLOWED_CONTENT_CHECKSUMS:
+        return None
+
+    return the_real_hashlib.new(name, *args, **kwargs)
+
+
+def init_and_validate(file, artifact_model, expected_digests):
+    """
+    Copied and updated (to comply with migrations) from pulpcore.
+    """
+    from django.conf import settings
+
+    digest_fields = []
+    for alg in ("sha512", "sha384", "sha256", "sha224", "sha1", "md5"):
+        if alg in settings.ALLOWED_CONTENT_CHECKSUMS:
+            digest_fields.append(alg)
+
+    if isinstance(file, str):
+        with open(file, "rb") as f:
+            hashers = {
+                n: hasher for n in digest_fields if (hasher := pulp_hashlib_new(n)) is not None
+            }
+            if not hashers:
+                return None
+
+            size = 0
+            while True:
+                chunk = f.read(1048576)  # 1 megabyte
+                if not chunk:
+                    break
+                for algorithm in hashers.values():
+                    algorithm.update(chunk)
+                size = size + len(chunk)
+    else:
+        size = file.size
+        hashers = file.hashers
+
+    mismatched_sha256 = None
+    for algorithm, expected_digest in expected_digests.items():
+        if algorithm not in hashers:
+            return None
+        actual_digest = hashers[algorithm].hexdigest()
+        if expected_digest != actual_digest:
+            # Store the actual value for later fixing if it differs from the package value
+            mismatched_sha256 = actual_digest
+
+    attributes = {"size": size, "file": file}
+    for algorithm in digest_fields:
+        attributes[algorithm] = hashers[algorithm].hexdigest()
+
+    return artifact_model(**attributes), mismatched_sha256
+
+
+def extract_wheel_metadata(filename):
+    """
+    Extract the metadata file content from a wheel file.
+    Return the raw metadata content as bytes or None if metadata cannot be extracted.
+    """
+    import zipfile
+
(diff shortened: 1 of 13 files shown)


pulp_python/app/tasks/sync.py
```diff
@@ -245,11 +249,28 @@ async def create_content(self, pkg):
...
+                        artifact = Artifact(sha256=md_sha256)
+
+                        metadata_artifact = DeclarativeArtifact(
+                            artifact=artifact,
+                            url=url,
+                            relative_path=f"{entry['filename']}.metadata",
+                            remote=self.remote,
+                            deferred_download=self.deferred_download,
+                        )
+                        d_artifacts.append(metadata_artifact)
```
gerrod3: One last thing we need to do is add the metadata_sha256 to the package since the PyPI json api doesn't have the value.

`package.metadata_sha256 = md_sha256`.

pulp_python/tests/functional/api/test_pypi_simple_api.py
```diff
@@ -4,21 +4,83 @@
...
+
+    proper, msgs = ensure_simple(
+        urljoin(distro.base_url, "simple/"),
+        {"shelf-reader": [PYTHON_WHEEL_FILENAME, PYTHON_EGG_FILENAME]},
+        sha_digests=PYTHON_XS_FIXTURE_CHECKSUMS,
+    )
+    assert proper, f"Simple API validation failed: {msgs}"
+
+    html_content = response.text
+    assert f'data-dist-info-metadata="sha256={PYTHON_WHEEL_METADATA_SHA256}' in html_content
```
gerrod3: Maybe we can add this check into ensure_simple. Pass in a new metadata_sha_digests dict that should check if certain links have this field set correctly.

pulp_python/tests/functional/api/test_sync.py
```diff
@@ -336,3 +338,21 @@ def test_sync_provenance(python_repo_with_sync, python_remote_factory, python_co
...
+def test_package_sync_with_metadata(
+    pulp_content_url,
+    python_distribution_factory,
+    python_remote_factory,
+    python_repo_with_sync,
+):
+    """
+    Test that the sync of a Python wheel package creates a metadata artifact.
+    """
+    remote = python_remote_factory(includes=PYTHON_XS_PROJECT_SPECIFIER)
```
gerrod3: We should choose a different package since this test runs in parallel meaning the upload tests will also be creating the same content. All the wheels on the fixtures should have the metadata file available.

pulp_python/app/migrations/0019_create_missing_metadata_artifacts.py
```diff
@@ -0,0 +1,140 @@
...
+            temp_metadata_path = temp_md.name
+            temp_md.write(metadata_content)
+            temp_md.flush()
+
+        metadata_artifact = artifact_model.init_and_validate(temp_metadata_path)
+        try:
+            metadata_artifact.save()
+        except IntegrityError:
+            metadata_artifact = artifact_model.objects.get(
+                sha256=metadata_artifact.sha256, pulp_domain=get_domain()
```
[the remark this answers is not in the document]
jobselko: I removed the part with `artifact_model.objects.get()` because if we do not have a metadata content artifact, we should not have a metadata artifact as well. So I am no longer using domains in the migration.

pulp_python/app/migrations/0019_create_missing_metadata_artifacts.py
```diff
@@ -0,0 +1,188 @@
...
+            metadata_digests = {"sha256": package.metadata_sha256}
+            metadata_artifact = artifact_to_metadata_artifact(
+                filename, main_artifact, metadata_digests, temp_dir, Artifact
+            )
+            if not metadata_artifact:
+                # Failed to build metadata artifact
+                continue
+
+            contentartifact = ContentArtifact(
+                artifact=metadata_artifact,
```
gerrod3: I don't think this will work. The metadata_artifact is just an instance in memory, it doesn't have a pk needed for this foreign key, and I don't believe that bulk_create goes and updates the passed in instances.

pulp_python/app/migrations/0019_create_missing_metadata_artifacts.py
```diff
@@ -0,0 +1,188 @@
...
+    else:
+        size = file.size
+        hashers = file.hashers
+
+    for algorithm, expected_digest in expected_digests.items():
+        if algorithm not in hashers:
+            return None
+        actual_digest = hashers[algorithm].hexdigest()
+        if expected_digest != actual_digest:
+            return None
```
gerrod3: Should we return something to tell the user that they will need to run the repair command?

pulp_python/app/migrations/0019_create_missing_metadata_artifacts.py
```diff
@@ -0,0 +1,211 @@
...
+            if result == "extraction_failed":
+                # Unset metadata_sha256 when metadata extraction fails
+                package.metadata_sha256 = None
+                packages_batch.append(package)
+                skipped_pkgs += 1
+                continue
+            if result is None:
+                # Failed to build metadata artifact (init_and_validate failed)
+                skipped_pkgs += 1
+                continue
```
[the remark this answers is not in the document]
jobselko: Updated

PR #1056, opened 2026-01-08
Draft: Return error msgs from ensure_simple correctly

jobselko:


pulp_python/tests/functional/utils.py
jobselko: `test_removed_content_not_published` (or maybe code?) contains a bug. I will look into it later

PR #1065, opened 2026-01-14
[PULP-1082] Add data-requires-python to Simple HTML API

### LINKED ISSUE #1054: Add data-requires-python to Simple HTML API
[PEP 503](https://peps.python.org/pep-0503/)

> A repository MAY include a data-requires-python attribute on a file link. This exposes the Requires-Python metadata field, specified in [PEP 345](https://peps.python.org/pep-0345/), for the corresponding release. Where this is present, installer tools SHOULD ignore the download when installing to a Python version that doesn’t satisfy the requirement. For example:
> 
> `<a href="..." data-requires-python="&gt;=3">...</a>`
> 
> In the attribute value, < and > have to be HTML encoded as `&lt;` and `&gt;`, respectively.

jobselko:
closes #1054


gerrod3 (approved):
Nice, good job catching the escaping behavior on the html template rendering.

PR #1186, opened 2026-04-09
Add more Pulp Exceptions.

aKlimau:
Assisted-by: Claude Sonnet 4.5 <noreply@anthropic.com>

### 📜 Checklist

- [x] Commits are cleanly separated with meaningful messages (simple features and bug fixes should be [squashed](https://pulpproject.org/pulpcore/docs/dev/guides/git/#rebasing-and-squashing) to one commit)
- [x] A [changelog entry](https://pulpproject.org/pulpcore/docs/dev/guides/git/#changelog-update) or entries has been added for any significant changes
- [x] Follows the [Pulp policy on AI Usage](https://pulpproject.org/help/more/governance/ai_policy/)
- [x] (For new features) - User documentation and test coverage has been added

### DIFF
--- a/CHANGES/+add-pulp-exceptions.feature
+++ b/CHANGES/+add-pulp-exceptions.feature
@@ -1,0 +1,1 @@
+Add more Pulp Exceptions.
--- a/pulp_python/app/exceptions.py
+++ b/pulp_python/app/exceptions.py
@@ -1,0 +1,54 @@
+from gettext import gettext as _
+
+from pulpcore.plugin.exceptions import PulpException
+
+
+class AttestationVerificationError(PulpException):
+    """
+    Raised when attestation verification fails.
+    """
+
+    error_code = "PYT0001"
+
+    def __init__(self, message):
+        super().__init__()
+        self.message = message
+
+    def __str__(self):
+        return f"[{self.error_code}] " + _("Attestation verification failed: {message}").format(
+            message=self.message
+        )
+
+
+class UnsupportedProtocolError(PulpException):
+    """
+    Raised when an unsupported protocol is used for syncing.
+    """
+
+    error_code = "PYT0002"
+
+    def __init__(self, protocol):
+        super().__init__()
+        self.protocol = protocol
+
+    def __str__(self):
+        return f"[{self.error_code}] " + _(
+            "Only HTTP(S) is supported for python syncing, got: {protocol}"
+        ).format(protocol=self.protocol)
+
+
+class InvalidAttestationsError(PulpException):
+    """
+    Raised when attestation data cannot be validated.
+    """
+
+    error_code = "PYT0003"
+
+    def __init__(self, message):
+        super().__init__()
+        self.message = message
+
+    def __str__(self):
+        return f"[{self.error_code}] " + _("Invalid attestations: {message}").format(
+            message=self.message
+        )
--- a/pulp_python/app/models.py
+++ b/pulp_python/app/models.py
@@ -412,16 +412,16 @@ class PythonRepository(Repository, AutoAddObjPermsMixin):
 
     def _check_for_package_substitution(self, new_version):
         """
-        Raise a ValidationError if newly added packages would replace existing packages that have
-        the same filename but a different sha256 checksum.
+        Raise a ValidationError if newly added packages would replace existing packages
+        that have the same filename but a different sha256 checksum.
         """
         qs = PythonPackageContent.objects.filter(pk__in=new_version.content)
         duplicates = collect_duplicates(qs, ("filename",))
         if duplicates:
             raise ValidationError(
-                "Found duplicate packages being added with the same filename but different checksums. "  # noqa: E501
-                "To allow this, set 'allow_package_substitution' to True on the repository. "
-                f"Conflicting packages: {duplicates}"
+                "Found duplicate packages being added with the same filename but different "
+                "checksums. To allow this, set 'allow_package_substitution' to True on the "
+                f"repository. Conflicting packages: {duplicates}"
             )
 
     def _check_blocklist(self, new_version):
--- a/pulp_python/app/serializers.py
+++ b/pulp_python/app/serializers.py
@@ -9,7 +9,8 @@ from django.db.utils import IntegrityError
 from drf_spectacular.utils import extend_schema_serializer
 from packaging.requirements import Requirement
 from packaging.version import InvalidVersion, Version
-from pydantic import TypeAdapter, ValidationError
+from pydantic import TypeAdapter
+from pydantic import ValidationError as PydanticValidationError
 from pypi_attestations import AttestationError
 from rest_framework import serializers
 
@@ -387,7 +388,7 @@ class PythonPackageContentSerializer(core_serializers.SingleArtifactContentUploa
                 attestations = TypeAdapter(list[Attestation]).validate_json(value)
             else:
                 attestations = TypeAdapter(list[Attestation]).validate_python(value)
-        except ValidationError as e:
+        except PydanticValidationError as e:
             raise serializers.ValidationError(_("Invalid attestations: {}").format(e))
         return attestations
 
@@ -654,7 +655,7 @@ class PackageProvenanceSerializer(core_serializers.NoArtifactContentUploadSerial
         try:
             provenance = Provenance.model_validate_json(data["file"].read())
             data["provenance"] = provenance.model_dump(mode="json")
-        except ValidationError as e:
+        except PydanticValidationError as e:
             raise serializers.ValidationError(
                 _("The uploaded provenance is not valid: {}").format(e)
             )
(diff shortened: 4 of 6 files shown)


CHANGES/+add-pulp-exceptions.feature
```diff
@@ -0,0 +1 @@
+Add more Pulp Exceptions.
```
jobselko, line 1: This is not a bug, but rather a feature.

pulp_python/app/tasks/sync.py
```diff
@@ -17,6 +17,7 @@
     Stage,
 )
 
+from pulp_python.app.exceptions import UnsupportedProtocolError
```
jobselko, line 25: Why do we need this? I do not remember that is was reported by CI. It also looks like sth that should be in pulpcore.

pulp_python/app/serializers.py
```diff
@@ -408,26 +408,23 @@ def deferred_validate(self, data):
         try:
             filename = data["relative_path"]
         except KeyError:
-            raise serializers.ValidationError(detail={"relative_path": _("This field is required")})
+            raise ValidationError(_("This field is required: relative_path"))
```
jobselko: Why do we need this? (not reported by CI)

pulp_python/app/serializers.py
```diff
@@ -408,26 +408,23 @@ def deferred_validate(self, data):
...
 
         artifact = data["artifact"]
         try:
             _data = artifact_to_python_content_data(filename, artifact, domain=get_domain())
         except ValueError:
-            raise serializers.ValidationError(
+            raise ValidationError(
```
jobselko: Why do we need this? (not reported by CI)

pulp_python/app/exceptions.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def __str__(self):
+        return (
+            f"[{self.error_code}] "
+            + _(
+                "Found duplicate packages being added with the same filename but different checksums. "  # noqa: E501
+            )
+            + _("To allow this, set 'allow_package_substitution' to True on the repository. ")
+            + _("Conflicting packages: {duplicates}").format(duplicates=self.duplicates)
+        )
```
jobselko: return f"[{self.error_code}] " + _(
            "Found duplicate packages being added with the same filename but different "
            "checksums. To allow this, set 'allow_package_substitution' to True on the "
            "repository. Conflicting packages: {duplicates}"
        ).format(duplicates=self.duplicates)

pulp_python/app/exceptions.py
```diff
@@ -0,0 +1,88 @@
...
+
+from pulpcore.plugin.exceptions import PulpException
+
+
+class ProvenanceVerificationError(PulpException):
+    """
+    Raised when provenance verification fails.
+    """
+
+    error_code = "PLPY0001"
```
jobselko: Do we already have a naming convention for plugins? Pulpcore uses "PLP", this introduces "PLPY"

aKlimau: We don't have a strict convention for plugins, but some plugins have already created exceptions with codes, so until we have a reason to make it all stricter we will go with it.

pulp_python/app/serializers.py
aKlimau: We need this to give more information to user/services team. As this exception can't contain any sensitive information we would rather display it to user for easier debugging (serializers.ValidationError will be sanitized)

jobselko: Discussed on the call that we might want to turn more exceptions into `PulpException`-based exceptions to avoid later sanitization in core 3.115.

pulp_python/app/exceptions.py
jobselko, line 1: `fetch_json_release_metadata` from `utils.py` can also raise an exception, this is called within the `repair` task

pulp_python/app/serializers.py
jobselko, line 562: This can be replaced by `InvalidPythonExtensionError` 

Edit: Actually not, we are inside `validate`, not `deferred_validate`

jobselko (changes requested):

pulp_python/app/tasks/sync.py
jobselko: Can we replace `UnsupportedProtocolError` by `SyncError`?

pulp_python/app/exceptions.py
jobselko: The name should be `PYT0001` (agreed in the core team mtg)

pulp_python/app/serializers.py
jobselko, line 395: This is called by `PythonPackageContentUploadSerializer.validate`, where it should raise a `serializers.ValidationError`

pulp_python/app/models.py
jobselko, line 439: `BlocklistedPackageError` instead of `ValidationError`

jobselko (changes requested):

pulp_python/app/serializers.py
aKlimau: It's still catching the same error as before the changes, I just renamed it from ValidationError to PydanticValidationError:
`from pydantic import TypeAdapter, ValidationError as PydanticValidationError`
Or am I getting smt wrong?

jobselko: `handle_attestations` now raises `AttestationVerificationError` instead of `serializers.ValidationError`

jobselko:
@aKlimau Can you please rebase?

jobselko:
I noticed that many of the pulp-exception changes were reverted. Could you share the reasoning behind this change?

aKlimau:
> I noticed that many of the pulp-exception changes were reverted. Could you share the reasoning behind this change?

As we decided to treat ValidationErrors and aiohttp ClientErrors as PulpExceptions there is now no need for new classes of PulpExceptions, that's why a bunch of them were removed and old ValidationErrors were brought back.

pulp_python/app/tasks/sync.py
```diff
@@ -115,7 +115,8 @@ async def run(self):
         url = self.remote.url.rstrip("/")
         downloader = self.remote.get_downloader(url=url)
         if not isinstance(downloader, HttpDownloader):
-            raise ValueError("Only HTTP(S) is supported for python syncing")
+            protocol = type(downloader).__name__
```
jobselko: `__name__` is a class name, not a protocol name. You can use this instead:
```
from urllib.parse import urlparse
urlparse(downloader.url).scheme
```

pulp_python/app/utils.py
```diff
@@ -357,7 +359,7 @@ def fetch_json_release_metadata(name: str, version: str, remotes: set[Remote]) -
             json_data = json.load(file)
         return json_data
     else:
-        raise Exception(f"Failed to fetch {url} from any remote.")
+        raise RemoteFetchError(url=url)
```
jobselko: Looking at this again, this is only called in the `repair` task where the exception is caught and discarded, so it never reaches the task. It is safe to drop this change. Sorry for confusion.

pulp_python/tests/functional/api/test_attestations.py
```diff
@@ -69,7 +69,7 @@ def test_verify_provenance(python_bindings, twine_package, python_content_factor
     with pytest.raises(PulpTaskError) as e:
         monitor_task(provenance.task)
     assert e.value.task.state == "failed"
-    assert "twine-6.2.0-py3-none-any.whl != twine-6.2.0.tar.gz" in e.value.task.error["description"]
+    assert "Provenance verification failed" in e.value.task.error["description"]
```
jobselko: This change is not needed, the serializers were not changed.

pulp_python/tests/functional/api/test_crud_content_unit.py
```diff
@@ -113,8 +113,7 @@ def test_content_crud(
     with pytest.raises(PulpTaskError) as e:
         response = python_bindings.ContentPackagesApi.create(**content_body)
         monitor_task(response.task)
-    msg = "The uploaded artifact's sha256 checksum does not match the one provided"
-    assert msg in e.value.task.error["description"]
+    assert "sha256 checksum does not match" in e.value.task.error["description"]
```
jobselko: This change is not needed, the serializers were not changed.

pulp_python/app/models.py
jobselko, line 1: Docstring changes are not needed, the exceptions were not changed.

pulp_python/app/exceptions.py
jobselko, line 1: Can you please change the exception numbers to 1-3 for `UnsupportedProtocolError`, `InvalidAttestationsError`, `AttestationVerificationError`?

pulp_python/app/tasks/sync.py
jobselko: `SyncError`

pulp_python/tests/functional/api/test_attestations.py
jobselko: This one is still present

jobselko (approved):

PR #1189, opened 2026-04-09
Fix repair_metadata OOM on large repositories

### LINKED ISSUE #1188: repair_metadata OOMs on large repositories (1000+ packages)
## Problem

When `repair_metadata` processes a repository with 1000+ packages, peak memory reaches ~7.7GB against an 8GB worker limit. The worker becomes unresponsive, misses its heartbeat, and Pulp marks the task as "Worker has gone missing." This consistently fails on the same large repos.

## Root Cause

Three factors compound to create the memory spike:

1. **`BULK_SIZE = 1000`** — batch and metadata_batch lists accumulate up to 1000 items before flushing
2. **Double wheel read** — each wheel is read from S3 twice: once in `artifact_to_python_content_data` and again in `artifact_to_metadata_artifact`
3. **No file handle cleanup** — `artifact.file` handles are never explicitly closed, keeping buffered data in memory

## Proposed Fix

- Reduce `BULK_SIZE` from 1000 to 250
- Reuse the temp file from the first wheel read for metadata artifact creation
- Explicitly close artifact file handles after each iteration

Expected peak memory reduction: from ~7.7GB to ~2-3GB for a 1042-package repo.

## Evidence

Task failure from production:
```json
{
  "state": "failed",
  "error": {"reason": "Worker has gone missing."},
  "progress_reports": [{"total": 1042, "done": 833}]
}
```

Prometheus metrics show memory spiking from 1.5GB to 7.7GB (96.8% of 8GB limit) during the repair task.

Related: PULP-1573

decko:
## Summary

- Reduce `BULK_SIZE` from 1000 to 250, flushing batches 4x more often to cap peak memory
- Eliminate double S3 read per wheel by reusing the temp file from metadata extraction for metadata artifact creation
- Explicitly close artifact file handles after each iteration to release S3 buffer memory

Fixes #1188

## Test plan

- [ ] Existing `test_repair.py` tests pass (metadata repair command, endpoint, artifact repair)
- [ ] New `test_metadata_repair_batch_boundary` passes with reduced BULK_SIZE
- [ ] Deploy to stage and run `repair-python-metadata.py --env stage --domain <large-domain>` to verify no OOM

JIRA: PULP-1573

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGES/1188.bugfix
+++ b/CHANGES/1188.bugfix
@@ -1,0 +1,1 @@
+Fixed "Worker has gone missing" errors during repair_metadata on large repositories (1000+ packages) by reducing peak memory consumption.
--- a/pulp_python/app/tasks/repair.py
+++ b/pulp_python/app/tasks/repair.py
@@ -1,4 +1,5 @@
 import logging
+import os
 from collections import defaultdict
 from gettext import gettext as _
 from itertools import groupby
@@ -8,18 +9,20 @@ from django.db.models import Prefetch
 from django.db.models.query import QuerySet
 from pulp_python.app.models import PythonPackageContent, PythonRepository
 from pulp_python.app.utils import (
-    artifact_to_metadata_artifact,
     artifact_to_python_content_data,
+    copy_artifact_to_temp_file,
+    extract_wheel_metadata,
     fetch_json_release_metadata,
+    metadata_content_to_artifact,
     parse_metadata,
 )
-from pulpcore.plugin.models import Artifact, ContentArtifact, ProgressReport
+from pulpcore.plugin.models import ContentArtifact, ProgressReport
 from pulpcore.plugin.util import get_domain
 
 log = logging.getLogger(__name__)
 
 
-BULK_SIZE = 1000
+BULK_SIZE = 250
 
 
 def repair(repository_pk: UUID) -> None:
@@ -118,11 +121,21 @@ def repair_metadata(content: QuerySet[PythonPackageContent]) -> tuple[int, set[s
                 .first()
                 .artifact
             )
-            new_data = artifact_to_python_content_data(package.filename, main_artifact, domain)
+            # Copy artifact to temp file once, extract both content data and metadata
+            temp_path = copy_artifact_to_temp_file(main_artifact, package.filename)
+            try:
+                new_data = artifact_to_python_content_data(
+                    package.filename, main_artifact, domain, temp_path=temp_path
+                )
+                metadata_content = (
+                    extract_wheel_metadata(temp_path) if package.filename.endswith(".whl") else None
+                )
+            finally:
+                os.unlink(temp_path)
             total_metadata_repaired += update_metadata_artifact_if_needed(
                 package,
                 new_data.get("metadata_sha256"),
-                main_artifact,
+                metadata_content,
                 metadata_batch,
                 pkgs_metadata_not_repaired,
             )
@@ -236,7 +249,7 @@ def update_package_if_needed(
 def update_metadata_artifact_if_needed(
     package: PythonPackageContent,
     new_metadata_sha256: str | None,
-    main_artifact: Artifact,
+    metadata_content: bytes | None,
     metadata_batch: list[tuple],
     pkgs_metadata_not_repaired: set[str],
 ) -> int:
@@ -248,7 +261,7 @@ def update_metadata_artifact_if_needed(
     Args:
         package: Package to check for metadata changes.
         new_metadata_sha256: The correct metadata_sha256 extracted from the main artifact, or None.
-        main_artifact: The main package artifact used to generate metadata.
+        metadata_content: Raw metadata bytes extracted from the wheel, or None.
         metadata_batch: List of tuples for batch processing (updated in-place).
         pkgs_metadata_not_repaired: Set of package PKs that failed repair (updated in-place).
 
@@ -265,13 +278,13 @@ def update_metadata_artifact_if_needed(
 
     # Create missing
     if not cas:
-        metadata_batch.append((package, main_artifact))
+        metadata_batch.append((package, metadata_content))
     # Fix existing
     elif new_metadata_sha256 != original_metadata_sha256:
         ca = cas.first()
         metadata_artifact = ca.artifact
         if metadata_artifact is None or (metadata_artifact.sha256 != new_metadata_sha256):
-            metadata_batch.append((package, main_artifact))
+            metadata_batch.append((package, metadata_content))
 
     if len(metadata_batch) == BULK_SIZE:
         not_repaired = _process_metadata_batch(metadata_batch)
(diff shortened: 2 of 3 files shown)


pulp_python/app/tasks/repair.py
```diff
@@ -118,18 +120,32 @@ def repair_metadata(content: QuerySet[PythonPackageContent]) -> tuple[int, set[s
...
-            new_data = artifact_to_python_content_data(package.filename, main_artifact, domain)
+            # For wheels, keep the temp file so we can reuse it for metadata artifact
+            is_wheel = package.filename.endswith(".whl")
+            if is_wheel:
+                new_data, temp_path = artifact_to_python_content_data(
+                    package.filename, main_artifact, domain, keep_temp_file=True
+                )
+            else:
+                new_data = artifact_to_python_content_data(package.filename, main_artifact, domain)
+                temp_path = None
```
jobselko: A better way would be to create a helper function that copies artifact to a temp file, then use the result for content data extraction in `artifact_to_python_content_data` and metadata extraction via `extract_wheel_metadata`. `update_metadata_artifact_if_needed` can then use temp file / metadata content in bytes instead of `main_artifact`.

pulp_python/app/utils.py
```diff
@@ -240,25 +240,43 @@ def compute_metadata_sha256(filename: str) -> str | None:
     return hashlib.sha256(metadata_content).hexdigest() if metadata_content else None
 
 
-def artifact_to_python_content_data(filename, artifact, domain=None):
+def artifact_to_python_content_data(filename, artifact, domain=None, keep_temp_file=False):
```
jobselko: There should be only minimal changes to use either temp path or the original logic with `with tempfile.NamedTemporaryFile`

pulp_python/app/utils.py
```diff
@@ -297,6 +315,36 @@ def artifact_to_metadata_artifact(
     return metadata_artifact
 
 
+def artifact_to_metadata_artifact_from_path(filename: str, temp_wheel_path: str) -> Artifact | None:
```
jobselko: This is not needed. `artifact_to_metadata_artifact` should take a new parameter (temp path or metadata content in bytes) and use either this or the original logic with `with tempfile.NamedTemporaryFile`.

jobselko (changes requested):

pulp_python/app/tasks/repair.py
decko: Agreed on the helper approach — will rework to have a small helper that copies the artifact to a temp file, then pass that path to both `artifact_to_python_content_data` and `artifact_to_metadata_artifact`.

One thing I'd like to discuss: passing metadata content in bytes through `update_metadata_artifact_if_needed` would couple the content extraction and artifact creation paths — `artifact_to_python_content_data` would need to know about and return the raw metadata bytes alongside the parsed data. Passing the temp file path keeps both functions independent: each opens the wheel for exactly what it needs. Would you be OK with just passing the temp path instead of bytes?

jobselko: The current implementation does not solve the issue with copying the same artifact twice. We could do something like this, since `artifact_to_metadata_artifact` does not need the whole artifact, but only a part of it - its metadata file:
```
temp_path = copy_artifact_to_temp_file(...)
try:
    new_data = artifact_to_python_content_data(...)
    metadata_content = (
        extract_wheel_metadata(temp_path) if package.filename.endswith(".whl") else None
    )
finally:
    os.unlink(temp_path)
total_metadata_repaired += update_metadata_artifact_if_needed(
    package, new_data.get("metadata_sha256"), metadata_content, ...
)
```
This way, we will copy the artifact only once, but the trade-off is running `extract_wheel_metadata` before we know we need it (i.e. before running `update_metadata_artifact_if_needed`). Would this change improve performance?

pulp_python/tests/functional/api/test_repair.py
```diff
@@ -331,3 +331,55 @@ def test_metadata_artifact_repair_endpoint(
     # Check deduplication
     assert len(main_artifact_hrefs) == 4
     assert len(metadata_artifact_hrefs) == 3
+
+
+def test_metadata_repair_batch_boundary(
```
jobselko: This test seems to be redundant and with a misleading name

pulp_python/app/utils.py
```diff
@@ -262,19 +281,26 @@ def artifact_to_python_content_data(filename, artifact, domain=None):
 
 
 def artifact_to_metadata_artifact(
```
jobselko, line 283: Changes made to this function are not used anywhere

pulp_python/app/tasks/repair.py
decko: Implemented. The loop now extracts `metadata_content` bytes via `extract_wheel_metadata(temp_path)` while the temp file exists, then deletes it immediately. The bytes are passed through the batch to a new `metadata_content_to_artifact` function that creates the Artifact from raw bytes.

The trade-off (running `extract_wheel_metadata` before knowing if repair is needed) is negligible — it's a zipfile read on an already-local temp file, no S3 involved.

pulp_python/tests/functional/api/test_repair.py
decko: Removed.

pulp_python/app/utils.py
decko: Reverted to original signature — removed the unused `temp_path` parameter. `artifact_to_metadata_artifact` now delegates to a new `metadata_content_to_artifact` for the bytes-to-Artifact step, which is also used by the repair batch.

jobselko:
Could you please squash the commits into one? Otherwise, it looks good!

decko:
> Could you please squash the commits into one? Otherwise, it looks good!

Done @jobselko. Just one commit.

jobselko (approved):
Thanks! I will keep this open, as @gerrod3 might want to review it as well. There are a few things that need small refactoring, but these are outside the scope of this PR.

CHANGES/1188.bugfix
```diff
@@ -0,0 +1,3 @@
+Reduced peak memory consumption of repair_metadata by lowering batch size from 1000 to 250,
+eliminating double S3 reads for wheel files, and closing artifact file handles after each
+iteration. This fixes "Worker has gone missing" errors on repositories with 1000+ packages.
```
gerrod3: Simplify this, too verbose. Keep it on one line.

gerrod3 (changes requested):
We can accept this change and backport it. In the future we will need to take a look at our util methods and refactor them, they've grown a bit absurd.

CHANGES/1188.bugfix
decko: Simplified to one line.

gerrod3 (approved):

PR #1271, opened 2026-07-14
[PULP-2012] Cache Simple API responses

### LINKED ISSUE #1254: Simple API responses should be cached server-side and served with HTTP cache headers
**Is your feature request related to a problem? Please describe.**

Pulp's Simple Repository API responses (both HTML and JSON) are regenerated from the database on every request and include no HTTP cache headers (`cache-control`, `etag`, `last-modified`). This means:

1. **Server-side**: Every request for a project's simple index page hits the database and regenerates the response, even though the content only changes when a new repository version is published.
2. **Client-side**: Without cache headers, clients like `uv` cannot cache responses locally and must re-fetch every package index page on every resolve.

By contrast, PyPI pre-generates and caches its simple index pages, and serves them with `cache-control: max-age=600, public` and an `etag`. This allows both the server to avoid redundant work and clients to skip the network entirely for 10 minutes, then do cheap `304 Not Modified` revalidation after that.

**Evidence**

```bash
# Pulp simple index — no cache headers
$ curl -sI "https://<pulp>/pulp/content/<domain>/pypi/public/simple/numpy/" | grep -iE 'cache|etag'
# (empty)

# PyPI — proper cache headers
$ curl -sI "https://pypi.org/simple/numpy/" | grep -iE 'cache|etag'
cache-control: max-age=600, public
etag: "5rCf07MprnFSJcmoYLKebQ"
```

**Impact**

| Scenario | PyPI | Pulp (no caching) |
|---|---|---|
| Single `uv pip compile` (cached) | 0.3 s | 8.3 s |
| Full lock-file refresh (18 dirs, 8× parallel) | ~30 s | 1 m 49 s |

**Describe the solution you'd like**

Two layers of caching:

1. **Server-side response caching**: Cache the generated Simple API responses (e.g. in Redis or on disk) so that repeated requests for the same project index page don't re-query the database and regenerate the response. The cache should be invalidated when a new repository version is published.

2. **HTTP cache headers**: Serve cached responses with `cache-control: max-age=600, public` and an `etag` derived from the repository version or publication timestamp. This enables client-side caching and cheap `304 Not Modified` revalidation, matching PyPI behavior.

Together, these would make repeated dependency resolution sub-second when packages haven't changed.

**Additional context**

- [PEP 503 – Simple Repository API](https://peps.python.org/pep-0503/)
- [PEP 691 – JSON Simple API](https://peps.python.org/pep-0691/)
- `uv` and `pip` both honor standard HTTP cache headers for index responses

jobselko:
closes #1254

### 📜 Checklist

### DIFF
--- a/pulp_python/app/cache.py
+++ b/pulp_python/app/cache.py
@@ -1,0 +1,32 @@
+from pulpcore.plugin.cache import CacheKeys, SyncContentCache
+from pulpcore.plugin.util import cache_key
+
+ACCEPT_HEADER_KEY = "accept_header"
+
+
+class PythonApiCache(SyncContentCache):
+    """
+    Cache for the Simple API.
+
+    Adds Accept header to the cache key so HTML and JSON responses are cached separately.
+    """
+
+    def __init__(self, base_key=None):
+        keys = (CacheKeys.path, CacheKeys.method, ACCEPT_HEADER_KEY)
+        super().__init__(base_key=base_key, keys=keys)
+
+    def make_key(self, request):
+        all_keys = {
+            CacheKeys.path: request.path,
+            CacheKeys.method: request.method,
+            ACCEPT_HEADER_KEY: request.headers.get("accept", ""),
+        }
+        return ":".join(all_keys[k] for k in self.keys)
+
+
+def find_base_path_cached(request, cached):
+    """
+    Resolve the distribution base_path for use as the Redis cache base_key.
+    """
+    path = request.resolver_match.kwargs["path"]
+    return cache_key(path)
--- a/pulp_python/app/pypi/views.py
+++ b/pulp_python/app/pypi/views.py
@@ -1,3 +1,4 @@
+import hashlib
 import logging
 from datetime import datetime, timedelta, timezone
 from itertools import chain
@@ -15,9 +16,11 @@ from django.http.response import (
     HttpResponseBadRequest,
     HttpResponseForbidden,
     HttpResponseNotFound,
-    StreamingHttpResponse,
 )
 from django.shortcuts import redirect
+from django.utils.decorators import method_decorator
+from django.views.decorators.cache import cache_control
+from django.views.decorators.http import condition
 from drf_spectacular.utils import extend_schema
 from dynaconf import settings
 from packaging.utils import canonicalize_name
@@ -31,6 +34,7 @@ from pulpcore.plugin.util import get_domain, get_url
 from pulpcore.plugin.viewsets import OperationPostponedResponse
 
 from pulp_python.app import tasks
+from pulp_python.app.cache import PythonApiCache, find_base_path_cached
 from pulp_python.app.models import (
     PackageProvenance,
     PythonDistribution,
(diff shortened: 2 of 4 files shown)


pulp_python/tests/functional/api/test_simple_cache.py
jobselko, line 1: I did not include tests for invalidation as this is functionality of pulpcore.

pulp_python/app/cache.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+
+    def __init__(self, base_key=None):
+        keys = (CacheKeys.path, CacheKeys.method, ACCEPT_HEADER_KEY)
+        super().__init__(base_key=base_key, keys=keys)
+
+    def __call__(self, func):
+        original = super().__call__(func)
+
+        def with_conditional_get(*args, **kwargs):
```
gerrod3: Did you check out the `condition` decorator from Django? https://docs.djangoproject.com/en/6.0/topics/conditional-view-processing/#the-condition-decorator

I think it can do most of the work that is being done here.

jobselko: I tried `conditional_page` (https://docs.djangoproject.com/en/5.2/topics/http/decorators/#django.views.decorators.http.conditional_page), but it did not work with DRF. I suspect `condition` may have a similar issue, but I will try it.

pulp_python/app/pypi/views.py
```diff
@@ -63,6 +66,30 @@
...
+def _add_cache_control_to_304(func):
+    """Ensure 304 responses include Cache-Control."""
+
+    def wrapper(*args, **kwargs):
+        response = func(*args, **kwargs)
+        if response.status_code == 304:
+            response["Cache-Control"] = CACHE_CONTROL
+        return response
+
+    return wrapper
```
gerrod3: We can replace this with the `cache_control` decorator from django as well.

https://docs.djangoproject.com/en/6.0/topics/http/decorators/#django.views.decorators.cache.cache_control

pulp_python/app/pypi/views.py
```diff
@@ -63,6 +66,30 @@
...
+
+
+def _etag_func(request, path, **kwargs):
+    """Compute unquoted ETag for the condition decorator. Returns None if no repo."""
+    distro = PyPIMixin.get_distribution(path)
```
gerrod3: This also needs to be inside the `try/except`

gerrod3 (changes requested):

pulp_python/app/cache.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    path = request.resolver_match.kwargs["path"]
+    base_key = cache_key(path)
+    if cached.exists(base_key=base_key):
+        return base_key
+    try:
+        distro = PythonDistribution.objects.get(base_path=path, pulp_domain=get_domain())
+    except ObjectDoesNotExist:
+        raise Http404(f"No PythonDistribution found for base_path {path}")
+    return cache_key(distro.base_path)
```
gerrod3:
```suggestion
    return cache_key(path)
```
I know this is partly copied from pulp_container, but I believe pulp_container is doing too much work. All we need is to convert the `base_path` from the request to the proper `base_key` that the cache is expecting. If it doesn't exist then Pulp will call the view's function and that will return the 404, no need to raise early. By simplifying this we remove 1 extra query for the case of the distro existing, but not yet being cached.

https://github.com/pulp/pulpcore/blob/main/pulpcore/cache/cache.py#L163

pulp_python/app/pypi/views.py
```diff
@@ -298,6 +313,9 @@ def get_provenance_url(self, package, version, filename):
         )
 
     @extend_schema(summary="Get index simple page")
+    @method_decorator(cache_control(max_age=600, public=True))
+    @method_decorator(condition(etag_func=_etag_func))
+    @PythonApiCache(base_key=find_base_path_cached)
```
gerrod3, line 318: Now this is really nice!

gerrod3 (changes requested):
Final changes, then LTGM

gerrod3 (approved):
Very nice, good work!

PR #1290, opened 2026-07-21
Use CentOS Stream 10 for CI

jobselko:
### 📜 Checklist

### DIFF
--- a/.github/workflows/scripts/before_install.sh
+++ b/.github/workflows/scripts/before_install.sh
@@ -55,7 +55,7 @@ pulp_scheme: "https"
 image:
   name: "pulp"
   tag: "ci_build"
-  ci_base: "ghcr.io/pulp/pulp-ci-centos9:latest"
+  ci_base: "ghcr.io/pulp/pulp-ci-centos10:latest"
   source: "${COMPONENT_SOURCE}"
   ci_requirements: $(test -f ci_requirements.txt && echo -n true || echo -n false)
   upperbounds: $(test "${TEST}" = "pulp" && echo -n true || echo -n false)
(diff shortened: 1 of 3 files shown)


dralley (approved):

dralley:
@jobselko `import distutils.util\nModuleNotFoundError: No module named 'distutils`

It was removed in python 3.12 https://docs.python.org/3/library/distutils.html

dralley:
I think it's not being used directly but rather indirectly, somehow.

dralley:
I think by https://pypi.org/project/pkginfo/

dralley:
Perhaps related, but pulp_python seems to have an undeclared dependency on `packaging`, which is only pulled in transitively but is used directly.

The logs seem to suggest that `packaging` is using `disutils`, but looking at their changelog, all such usages were removed long ago (2021). Maybe an ancient version is being pulled in somehow.

dralley:
@jobselko Is this an actual fix or does it just avoid the issue being triggered?

jobselko:
@dralley It just avoids triggering the issue, but I wanted to try it. I will add packaging as a dependency.

dralley:
You could pass `--target` to pip to make sure the packages don't get installed to the local site-packages, which seems to be the issue.  The pull-through test runs `pip install` against a package into Pulp's own local environemnt, which installs an ancient version of `packaging` as a dependency.  `pulp_python` doesn't conflict because the dependency on `packaging` isn't declared in the first place.


>   -t, --target `<dir>`          

> Install packages into `<dir>`. By default this will not replace existing files/folders in `<dir>`. Use --upgrade to replace existing packages in `<dir>` with new  versions.

jobselko:
`--no-deps` fix is sufficient to unblock the CI and is simpler than `--target`. We can add the `packaging` dependency in a follow-up once we determine the correct lower bound, which will require some investigation.

dralley:
Simpler, yes, but tests installing packages into the same environment the code is running in is still something to avoid

PR #1291, opened 2026-07-21
Use CentOS Stream 10 for CI

dralley:
And update template


dralley:
Beaten to the punch by @jobselko

PR #1327, opened 2026-08-13
[PULP-2199] Add SLSA attestation verification support

jobselko:
### 📜 Checklist


jobselko:
The first commit contains the applied patch and tests, the second covers the changes needed to upstream it. I will squash them at the end.

.ci/assets/keys/test-key-private.pem
gerrod3, line 1: I think you need to add this to a git secrets file so that the scanner doesn't flag our repo as leaking secrets.

pulp_python/app/provenance.py
```diff
@@ -58,14 +92,121 @@ class Provenance(BaseModel):
...
+
 def verify_provenance(filename, sha256, provenance, offline=True):
-    """Verify the provenance object is valid for the package."""
+    """Verify the provenance object is valid for the package.
+
+    Attestations with valid Sigstore certificates are verified through the
+    standard Sigstore path. Attestations without certificates are verified
+    against a custom public key configured via ATTESTATION_VERIFICATION_KEY.
+    Currently, it supports RSA PKCS1v15 signatures and SLSA v0.2 provenance
+    publisher enrichment.
```
gerrod3, line 190: The publisher enrichment is technically a side-effect of the method. Not sure it's the best thing to copy.

pulp_python/app/provenance.py
```diff
@@ -58,14 +92,121 @@ class Provenance(BaseModel):
...
+                policy = publisher._as_policy()
+                sig_bundle = attestation.to_bundle()
+                checkpoint = sig_bundle.log_entry._inner.inclusion_proof.checkpoint
+                staging = "sigstage.dev" in checkpoint.envelope
+                attestation.verify(policy, dist, staging=staging, offline=offline)
+            else:
+                stmt = _verify_statement_subject(attestation, dist)
+                _enrich_publisher_from_statement(stmt, publisher)
+                if verification_key:
+                    _verify_signature(attestation, verification_key)
```
gerrod3, line 207: Only one key to verify against across all repositories and all domains. I wonder if we could do something better. What services has is custom built for their use-case, so I don't really want to just move it over as is.

gerrod3 (commented):
I'm having second thoughts about moving this patch over. I think we should spend some more time thinking about the feature and what would be useful to more people.

PR #1351, opened 2026-08-25
Fix http error reporting and json metadata handling

hstct:
Bandersnatch masks http errors when syncing packages as stale metadata instead of reporting them directly the `raise_for_status` flag will show the actual error like if the package is not here 404, or there is a server side error 5xx. This helps debugging. It also doesn't run into a retry loop but fails fast.

The other change allows native json list values to be allowed when rendering python package metadata, an error that previously was masked as stale metadata.

## Summary by CodeRabbit

* **Bug Fixes**
  * Python synchronization now reports HTTP errors accurately instead of misclassifying them as stale metadata.
  * Package metadata containing native JSON lists is now handled correctly.

### DIFF
--- a/CHANGES/+fixed-error-reporting-and-json-metadata-handling.bugfix
+++ b/CHANGES/+fixed-error-reporting-and-json-metadata-handling.bugfix
@@ -1,0 +1,1 @@
+Fixed python syncs masking HTTP errors as stale metadata and handling native json list values in package metadata
--- a/pulp_python/app/tasks/sync.py
+++ b/pulp_python/app/tasks/sync.py
@@ -130,6 +130,7 @@ class PythonBanderStage(Stage):
                 auth=downloader.auth,
                 proxy=downloader.proxy,
                 proxy_auth=downloader.proxy_auth,
+                raise_for_status=True,
             )
 
             deferred_download = self.remote.policy != Remote.IMMEDIATE
--- a/pulp_python/app/utils.py
+++ b/pulp_python/app/utils.py
@@ -441,7 +441,7 @@ def json_to_dict(data):
         dictionary: of JSON string
 
     """
-    if isinstance(data, dict):
+    if isinstance(data, (dict, list)):
         return data
 
     return json.loads(data)


gerrod3:
I don't think this is the right solution and I don't think you have identified what is really going on. If you have logs that would be helpful.

hstct:
@gerrod3 Thanks. I agree that the origin of the mixed json representation isn't identified yet.

What I was able to confirm from the affected system is the following:

- The Proxy sync failed while requesting `/pypi/<package>/json` from the Katello Pulp instance.
- Without `raise_for_status=True`, the 500 HTTP returned by `pulpcore-content` was being misreported as`stale-metadata error, which sent the investigation in the wrong direction.
- The traceback for affected packages was like:
```
python_content_to_json(...)
  -> python_content_to_info(latest_content[0])
  -> json_to_dict(content.provides_extras)
  -> json.loads(data)

TypeError: the JSON object must be str, bytes or bytearray, not list
```
- On the affected system the relevand db values looked like:
```
requires_dist = "[...]"
provides_dist = "[]"
dynamic = []
license_file = []
provides_extras = []
```

so some json fields contained serialized json strings while others contained native json arrays.

I reproduced that state on a test instance by changing `provides_extras` from the json string `"[]"` to the json array `[]`. With the current implementation, the `/pypi/<package>/json` endpoint raises the same `TypeError`. Allowing an already deserialized list in `json_to_dict()` makes the same endpoint work.

I have not identified what originally caused those fields to be stored as native arrays on the affected system. We weren't able to reproduce that through the normal sync/upload paths, and those paths currently call `json.dumps()` for these fields. So I don't want to claim that this patch fixes the underlying source of the mixed representation.

The reason I proposed the reader side change is that these are `JSONField`, and `json_to_dict()` already explicitly accepts an already deserialized `dict` but not an already deserialized `list`. The patch makes that handling consistent and prevents the metadata endpoint from returning 500 for this observed db state.

This proposed change did however unblock the affected system by allowing pulp to serve the package metadata again.

For reference the proxy side error was:

```
aiohttp.client_exceptions.ClientResponseError: 500, message='Internal Server Error'
URL: .../pulp/content/.../pypi/bcrypt/json
```

and the corresponding content side failure was:

```
python_content_to_info(...)
"provides_extras": json_to_dict(content.provides_extras) or None

json_to_dict(...)
return json.loads(data)

TypeError: the JSON object must be str, bytes or bytearray, not list
```

PR #1359, opened 2026-08-27
feat: add HTTP catalog for Python packages and repository metrics

### LINKED ISSUE #1358: Extension of Python REST API (repo index, metrics, content filters)
Lightwell UI lists Python packages, versions, and repository counts through content-sources-backend. That backend reads Pulp Postgres directly (Tangy). We need Pulp HTTP APIs so the path is UI → backend → Pulp REST, not SQL.

Today `GET …/content/python/packages/` is one row per file (wheel + sdist). It cannot paginate distinct package names or return versions per package.

`{pk}` is the repository UUID. Latest complete repository version is implied.

| UI path | Pulp |
|---------|------|
| PackageList | **New** `GET …/repositories/python/python/{pk}/packages/` — one row per `name_normalized`, paginate, `name_normalized__istartswith` |
| PackageVersions | **Extend** existing `GET …/content/python/packages/` with `collapse_builds=true` and serializer field `base_version` (still use `packagetype=sdist`) |
| PackageGet | Existing content list (`name` + `version` + `packagetype=sdist`); add `base_version` on the serializer. Omit `collapse_builds`. |
| RepoMetrics | **New** `GET …/repositories/python/python/{pk}/metrics/` — `package_count`, `version_count`, `build_count` |

`collapse_builds=true` (default false): strip trailing `\.[a-zA-Z]+-\d+$` from `version`, one row per `(name_normalized, base_version)`.

PackageList rows: `name`, `name_normalized`, `versions[]`, `latest_releases[]` (`version`, `created_at`, `release`). Same version keys in both lists. Pulp-native shape, not Tangy DTOs.

TenSt:
### 📜 Checklist

- [x] Commits are cleanly separated with meaningful messages (simple features and bug fixes should be [squashed](https://pulpproject.org/pulpcore/docs/dev/guides/git/#rebasing-and-squashing) to one commit)
- [x] A [changelog entry](https://pulpproject.org/pulpcore/docs/dev/guides/git/#changelog-update) or entries has been added for any significant changes
- [x] Follows the [Pulp policy on AI Usage](https://pulpproject.org/help/more/governance/ai_policy/)
- [x] (For new features) - User documentation and test coverage has been added

This PR:
- Adds repository `packages` and `metrics` HTTP endpoints for catalog clients
- Adds `last_updated`, `ordering`, and `name_normalized` prefix/substring search on the package catalog (minimum 3 characters; GIN trigram index)
- Returns catalog `versions` and `latest_releases` newest-first (PEP 440)
- Adds `collapse_builds` and `base_version` on the Python content API

Closes #1358

## Summary by CodeRabbit

- **New Features**
  - Added REST API endpoints for browsing repository package catalogs and viewing package, version, and build metrics.
  - Added pagination, package-name filtering and search, ordering, repository-version selection, and optional rebuild collapsing.
  - Added a read-only `base_version` field for Python package content.
  - Added access controls for the new catalog and metrics actions.

- **Documentation**
  - Added catalog browsing guides and documented the new API capabilities in the changelog.


pulp_python/app/catalog.py
```diff
@@ -0,0 +1,166 @@
...
+    """Python package content contained in ``repository_version``."""
+    if repository_version is None:
+        return PythonPackageContent.objects.none()
+    return PythonPackageContent.objects.filter(pk__in=repository_version.content)
+
+
+def apply_package_prefix_filters(queryset, name_normalized_prefix=None, name_prefix=None):
+    """Apply case-insensitive prefix filters used by the package index."""
+    if name_normalized_prefix:
+        queryset = queryset.filter(name_normalized__istartswith=name_normalized_prefix)
```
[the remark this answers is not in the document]
TenSt: The `name_normalized__istartswith` expects that input should be already normalized, I don't think that backend should be responsible for that.

pulp_python/app/viewsets.py
```diff
@@ -247,6 +281,169 @@ def sync(self, request, pk, **kwargs):
...
+        names_qs = distinct_package_names_qs(
+            content_qs, repository, repo_version, ordering=ordering
+        )
+        page = self.paginate_queryset(names_qs)
+        rows = assemble_package_index(
+            content_qs,
+            page if page is not None else list(names_qs),
+            repository,
+            repo_version,
+        )
```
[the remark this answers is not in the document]
TenSt: Releases are empty - this is intentional until we will start having rebuilds. If release is empty, it means there is no rebuild and original version (1.0.0) should be used.

pulp_python/tests/unit/test_catalog.py
```diff
@@ -0,0 +1,119 @@
...
+@pytest.mark.parametrize(
+    "value,expected",
+    [
+        (None, None),
+        ("", None),
+        ("   ", None),
+        ("djan", "djan"),
+        ("DJAN", "djan"),
+        ("JAN", "jan"),
+        ("  HTTP ", "http"),
```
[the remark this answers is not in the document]
TenSt: This is intentional.

CHANGES/1358.feature
```diff
@@ -0,0 +1 @@
+Added repository package catalog and metrics endpoints, plus ``collapse_builds`` and ``base_version`` on the Python package content API. The catalog includes ``last_updated``, ``ordering``, newest-first PEP 440 ``versions``/``latest_releases``, and ``name_normalized`` prefix/substring search (at least 3 characters). A trailing rebuild suffix is ``\.[a-zA-Z]+-[^.]+$`` (for example ``5.3.17.rhlw-00001-n0001`` groups with ``5.3.17``). ``latest_releases[].release`` is that suffix on the newest unit in the group, or empty when the stored version has none. Existing installs pick up access policy for the new actions on migrate unless the policy was customized.
```
gerrod3, line 1: Keep the first sentence, remove the rest.

docs/user/guides/catalog.md
```diff
@@ -0,0 +1,116 @@
+# Browse the package catalog
+
+Pulp CLI commands for these endpoints are generated from the OpenAPI spec in a separate package; until that is updated, use HTTP.
```
gerrod3, line 3: AI-ahh comment. Remove

pulp_python/app/catalog.py
```diff
@@ -0,0 +1,215 @@
...
+def collapse_python_builds(queryset):
+    """Keep one content unit per ``(name_normalized, base_version)``.
+
+    ``base_version`` is ``version`` with a trailing rebuild suffix stripped.
+    The unit with the latest ``pulp_created`` is kept. Callers that want one
+    row per logical version (not per wheel/sdist) should also filter
+    ``packagetype``.
+    """
+    return (
+        queryset.prefetch_related(None)
```
gerrod3, line 43: Why are we throwing away anything that has been prefetched? There are things that pulpcore pulls in that would be bad to just lose.

pulp_python/app/versions.py
```diff
@@ -0,0 +1,105 @@
...
+CI unit tests run with ``-p no:pulpcore``, so this module must stay importable
+before Django apps are loaded.
+"""
+
+import re
+
+from packaging.version import InvalidVersion, Version
+
+# Last dot-segment is a rebuild if it is letters, dash, rest of that segment.
+# POSIX string shared with SQL REGEXP_REPLACE. Not hard-coded to "rhlw".
```
gerrod3, line 12: remove the hard-coded comment.

pulp_python/app/viewsets.py
```diff
@@ -247,6 +282,179 @@ def sync(self, request, pk, **kwargs):
...
+        description=(
+            "Return one row per distinct package name in a repository version "
+            "(latest complete version if repository_version is omitted). "
+            "Pagination count is the number of distinct packages, not files. "
+            "Each row includes last_updated (newest membership among any rebuild), "
+            "versions (logical version keys after rebuild-suffix strip, newest first), "
+            "and latest_releases (newest rebuild per logical version, same order). "
+            "set(versions) === set(latest_releases[].version)."
+        ),
+        parameters=[
```
gerrod3, line 296: Is there not a better way to define the filter parameters for this endpoint then explicitly declaring them in `extend_schema`. Could we use `get_filterset` https://django-filter.readthedocs.io/en/latest/guide/rest_framework.html#overriding-filterset-creation?

pulp_python/app/viewsets.py
```diff
@@ -539,6 +766,18 @@ class PythonPackageSingleArtifactContentUploadViewSet(
     minimal_serializer_class = python_serializers.MinimalPythonPackageContentSerializer
     filterset_class = PythonPackageContentFilter
 
+    def filter_queryset(self, queryset):
+        """Apply ``collapse_builds`` after other backends so DISTINCT ON stays valid."""
+        queryset = super().filter_queryset(queryset)
+        if getattr(self, "action", "") != "list":
+            return queryset
+        raw = self.request.query_params.get("collapse_builds")
```
gerrod3, line 774: We should put this logic in the PythonPackageContentFilter. See https://github.com/carltongibson/django-filter/blob/main/django_filters/filterset.py#L224

CLAUDE.md
```diff
@@ -58,3 +58,7 @@ When patchback fails to cherry-pick a PR into an older branch, you need to manua
...
...jango-backed modules (``pulp_python.app.utils``, ``catalog``, models, viewsets). Keep ``strip_build_suffix``, ``BUILD_SUFFIX_PATTERN``, ``version_sort_key``, ``normalize_package_index_ordering``, and ``normalize_name_normalized_search`` in ``pulp_python/app/versions.py``. The rebuild suffix is the last dot-segment matching POSIX ``\.[a-zA-Z]+-[^.]+$`` (letters, dash, rest of that segment; not hard-coded to ``rhlw``). Python ``re`` and SQL ``REGEXP_REPLACE`` share ``BUILD_SUFFIX_PATTERN``; ``catalog.py`` may import it. Catalog ``latest_releases`` keeps the newest ``pulp_created`` unit per logical version; ``release`` is ``rebuild_release`` of that stored ``version`` (empty when there is no suffix). Catalog ``name_normalized`` prefix/substring filters lowercase the input, use ``LIKE`` (not ``ILIKE``) against the trigram GIN index, and reject values shorter than 3 characters. Simple-index ``DISTINCT ON (name_normalized)`` must ``ORDER BY name_normalized, name`` so the displayed project name is deterministic when metadata names differ (``msg-parser`` vs ``msg_parser``). Without the secondary sort, ``ensure_simple`` can miss the ``msg-parser`` link even though both files were published.
```
gerrod3, line 64: Remove this.

gerrod3 (changes requested):
I haven't gone deep yet, I'll try to find time to review more closely. Can you replace all double ticks with single ticks?

docs/user/guides/catalog.md
mdellweg, line 3:
```suggestion
```
This promise is helping no user.

For the rest of this file, please write for the target audience, a user. Do not add unrelated or irrelevant statements. Do not embark on implementation technicalities, but focus on why a user should do what. Give examples and explain them and their output. If using a special syntax, say so.

pulp_python/app/models.py
```diff
@@ -246,6 +247,13 @@ def __str__(self):
     class Meta:
         default_related_name = "%(app_label)s_%(model_name)s"
         unique_together = ("sha256", "_pulp_domain")
+        indexes = [
+            GinIndex(
```
mdellweg, line 251: What exactly is this index helping with?

pulp_python/app/versions.py
```diff
@@ -0,0 +1,105 @@
...
+NAME_NORMALIZED_SEARCH_MIN_LENGTH = 3
+
+
+def strip_build_suffix(version):
+    """Return ``version`` with a trailing rebuild suffix removed, else unchanged.
+
+    A rebuild is the last dot-segment matching ``BUILD_SUFFIX_PATTERN``.
+    """
+    if not version:
+        return version
```
mdellweg, line 27: Do we really expect bad data all the way deep down in the stack?
Can't we guard against that once when digesting a version?
Can a python package have an invalid version in the first place?

pulp_python/app/versions.py
```diff
@@ -0,0 +1,105 @@
...
+before Django apps are loaded.
+"""
+
+import re
+
+from packaging.version import InvalidVersion, Version
+
+# Last dot-segment is a rebuild if it is letters, dash, rest of that segment.
+# POSIX string shared with SQL REGEXP_REPLACE. Not hard-coded to "rhlw".
+BUILD_SUFFIX_PATTERN = r"\.[a-zA-Z]+-[^.]+$"
```
mdellweg, line 13: As far as I know python and postgres regular expression syntax is different.

pulp_python/app/versions.py
```diff
@@ -0,0 +1,105 @@
...
+def rebuild_release(version):
+    """Return the rebuild qualifier without the leading dot, or an empty string."""
+    if not version:
+        return ""
+    base = strip_build_suffix(version)
+    if version == base:
+        return ""
+    if version.startswith(base + "."):
+        return version[len(base) + 1 :]
+    return ""
```
mdellweg, line 40: What just happened when we land on this return?

pulp_python/app/versions.py
```diff
@@ -0,0 +1,105 @@
...
+    """
+    if not version:
+        return (-1, "")
+    try:
+        return (0, Version(version))
+    except InvalidVersion:
+        return (-1, str(version))
+
+
+def normalize_package_index_ordering(raw_values):
```
mdellweg, line 56: This looks like a reimplementation of (a significant part of) django-filters.

pulp_python/tests/unit/test_catalog.py
```diff
@@ -0,0 +1,135 @@
...
+        ("1.0.0.ABC-99", "1.0.0"),
+        ("1.0.foo-bar", "1.0"),
+        ("1.0.rhlw-١", "1.0"),
+        ("4.3.0-redhat-1", "4.3.0-redhat-1"),
+        ("5.3.18-anything", "5.3.18-anything"),
+        ("5.3.18.anything", "5.3.18.anything"),
+        ("1.0.rhlw-00003.extra", "1.0.rhlw-00003.extra"),
+        ("1.0.rhlw-", "1.0.rhlw-"),
+        ("", ""),
+        (None, None),
```
mdellweg, line 40: What are the actual rules here? It is not apparent from the examples.

Also it would really be nice to keep the versions pep440 compatible:
https://packaging.python.org/en/latest/specifications/version-specifiers/#local-version-identifiers

pulp_python/app/viewsets.py
TenSt: Moved into `PythonPackageContentFilter.filter_queryset`

pulp_python/app/models.py
TenSt: It backs catalog `name_normalized` prefix/substring search (LIKE / gin_trgm_ops). Exact lookups still use the btree on `name_normalized`

pulp_python/app/viewsets.py
TenSt: Catalog params are now `PythonRepositoryPackageFilter`. We apply it inside `packages()` because this viewset’s queryset is repositories. limit/offset remain declared for OpenAPI.

pulp_python/app/catalog.py
TenSt: Changed logic a bit, but as far as I understood, the issue is in how Django handles `DISTINCT ON` with pre-fetches. Here is AI explanation:
`
DISTINCT ON cannot reuse pulpcore’s list prefetch (contentartifact_set): Django clones that queryset, and the JOIN/lookups fight the DISTINCT ON/ORDER BY. We drop the pending prefetch, collapse, then prefetch contentartifact_set again for the reduced PK set so the serializer still gets artifacts in one query.
`

Please let me know if I understood something incorrectly (I don't have a lot of experience with Django) and you have a better/cleaner solution for this.

pulp_python/app/versions.py
TenSt: There is no guard check during ingestion. We take the version that is provided or put an empty string. Also, no validation of PEP 440 is present. So at this point in code, empty string or any string is allowed. We just assume that it would be of proper format and not empty. Adding/changing ingestion guard is out of the scope of this PR.

pulp_python/app/versions.py
TenSt: This particular pattern is the same for both python and postgres. I will change it to be PEP 440 compatible (as this is what we expect it to be too). This one was taken from maven version format (which is different) and was not adapted yet.

pulp_python/app/versions.py
TenSt: This is unreachable situation in this case. Removing it.

pulp_python/tests/unit/test_catalog.py
TenSt: Yes, good catch. These are taken from maven format (which does not apply to python). I'll be changing that using pep440 versions.

pulp_python/app/versions.py
TenSt: `normalize_package_index_ordering` only does: take `?ordering=…`, check it is one of `name / name_normalized / last_updated`, then add those extra keys so the grouped list paginates stably.

django-filter still checks that the query param is allowed. It cannot apply the sort itself, because the sort is not on the model queryset.

pulp_python/tests/unit/test_catalog.py
TenSt: Updated it to 440pep version style.

TenSt:
Hi @gerrod3 @mdellweg! Thank you for the review.
I've addressed all of the comments and added 2 more commits to make it easier to review and track changes. I'll squash them all into 1 commit once you'll be ready to approve.

PR #1375, opened 2026-09-09
fix: RSS feeds show stale data for rebuilt packages

### LINKED ISSUE #1374: RSS feeds show stale timestamps for rebuilt packages and have a low update limit
## Problem

The RSS feeds introduced in #1320 have two issues:

1. **Stale timestamps for rebuilds**: `iter_releases()` and `iter_projects()` use `Min("file_added_at")` to aggregate timestamps. When new files are added for an existing `(name, version)` pair (e.g., a rebuild with a different PEP 427 build tag), the feed continues to show the original timestamp. The `<guid>` also stays the same, so RSS readers never surface the rebuild as a new item.

2. **Low update limit**: `UPDATES_LIMIT` is 100, which is too small for indexes that receive batch promotions of 100+ packages at once. Items beyond the limit are silently truncated.

## Proposed Fix

- Change `Min` to `Max` on all aggregations so the feed reflects the most recently added file.
- Append a `#timestamp` fragment to the `<guid>` so RSS readers treat each rebuild as a new item. Set `isPermaLink="false"` accordingly.
- Raise `UPDATES_LIMIT` from 100 to 500.

## Reproducer

```bash
# Feed only shows build-tag -1 (Sep 5), missing -2 (Sep 8):
curl -s '.../rss/project/ctranslate2/releases.xml' \
  | python3 -c "import sys,xml.etree.ElementTree as ET; [print(f'{i.findtext(\"title\")} — {i.findtext(\"pubDate\")}') for i in ET.fromstring(sys.stdin.read()).findall('.//item')]"

# Pulp content API shows both build tags:
curl -s '.../api/v3/content/python/packages/?name=ctranslate2&version=4.8.2&limit=5&fields=filename,pulp_created&ordering=-pulp_created' \
  | jq -r '.results[] | "\(.filename) — \(.pulp_created[:19])"'
```

ryanpetrello:
## Summary

- Use `Max()` instead of `Min()` for the `added_at` timestamp aggregation so feeds reflect the most recently added file for each `(name, version)` pair
- Append a `#timestamp` fragment to the `<guid>` so RSS readers treat rebuilds as new items, and set `isPermaLink="false"` accordingly
- Raise `UPDATES_LIMIT` from 100 to 500 for indexes with large batch promotions

## Context

When a package is rebuilt with a different PEP 427 build tag (e.g., `pkg-1.0-1-...whl` then `pkg-1.0-2-...whl`), the RSS feed continued to show the original timestamp and guid because `iter_releases()` used `Min("file_added_at")`. RSS readers never surfaced the rebuild as a new item because the `<guid>` was unchanged.

The 100-entry limit in `updates.xml` was also too small for indexes that receive batch promotions of 100+ packages at once (we've seen promotions of up to 117 packages in a single hour).

Fixes #1374

### DIFF
--- a/CHANGES/1374.bugfix
+++ b/CHANGES/1374.bugfix
@@ -1,0 +1,1 @@
+RSS feeds now reflect the most recently added file for each release, produce unique guids when content is updated, and support up to 500 entries in `updates.xml`.
--- a/pulp_python/app/pypi/feeds.py
+++ b/pulp_python/app/pypi/feeds.py
@@ -2,7 +2,7 @@ import re
 from email.utils import getaddresses
 from urllib.parse import urljoin
 
-from django.db.models import F, FilteredRelation, Min, Q
+from django.db.models import F, FilteredRelation, Max, Min, Q
 from django.http.response import HttpResponse, HttpResponseNotFound
 from django.utils.decorators import method_decorator
 from django.utils.feedgenerator import Rss201rev2Feed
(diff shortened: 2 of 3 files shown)


gerrod3 (approved):
LGTM

PR #1380, opened 2026-09-14
we default to json respond even when html needed

YasenT:
Right now, when a standard browser tries to open the simple endpoint, it gets a json response instead of the expected html one. This was raised by users.
Pushing forward a quick fix

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved PyPI Simple API content negotiation using `Accept` header quality values and media-type specificity.
  - Correctly respects explicit exclusions (`q=0`) and wildcard preferences when selecting HTML or JSON responses.
  - Added support for `format` and `fmt` query parameters on list and retrieve requests.
  - Improved fallback behavior when requested response formats cannot be served.

- **Tests**
  - Expanded coverage for weighted, wildcard, and excluded media types in `Accept` headers.


gerrod3:
I can't reproduce this issue on a local pulp-instance. Are you sure the problem isn't with your deployment?

PR #1382, opened 2026-09-15
fix: stabilize RSS feed GUIDs for multi-arch wheel publication

### LINKED ISSUE #1381: RSS feed GUIDs are unstable when multi-architecture wheels arrive at different times
## Summary

The GUID fix from #1374 (appending `#timestamp` to make rebuilds visible) causes feed readers to show duplicate entries when platform-specific wheels for the same `(name, version, build_tag)` arrive at different times.

## Root cause

`iter_releases()` groups by `(name_normalized, version)` and annotates `added_at=Max("file_added_at")`. The GUID is `link#pubdate.isoformat()`. When a new architecture wheel is promoted (e.g., s390x arriving an hour after x86_64/aarch64/ppc64le), `Max(file_added_at)` shifts, producing a different GUID. Feed readers treat the new GUID as a new item.

## Proposed fix

Replace the timestamp-based GUID fragment with a build-tag-based fragment. Use `ArrayAgg("filename")` in the query, extract PEP 427 build tags from wheel filenames in Python, and construct GUIDs like `link#builds=1,2`.

This makes GUIDs:
- **Stable** when additional platform wheels arrive for the same build tag
- **Different** when a genuine rebuild introduces a new build tag (the original #1374 requirement)

ryanpetrello:
## Summary

- RSS feed GUIDs now use PEP 427 build tags instead of timestamps, preventing duplicate entries in feed readers when platform wheels for the same build arrive at different times.
- GUIDs still change when a new build tag is introduced (the #1374 requirement).

Fixes #1381

## Test plan

- [x] Existing feed tests updated and passing
- [x] New functional test: build tag 1 -> GUID `#builds=1`; add tag 2 -> GUID `#builds=1,2`; add new arch for tag 2 -> GUID stable
- [x] Unit tests for `_build_tag_fragment`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * RSS feed entry identifiers now remain stable when additional platform-specific wheels are published for an existing release.
  * Prevented duplicate feed entries for multi-architecture builds sharing the same build tag.
  * Feed identifiers now change only when a new PEP 427 build tag is introduced.
  * Feed publication dates continue to update when additional files are published without changing the build tag.


pulp_python/app/pypi/feeds.py
```diff
@@ -91,14 +95,30 @@ def iter_projects(content, repo_ver, limit=PACKAGES_LIMIT):
...
+    or an empty string for sdists and wheels without build tags.
+    """
+    tags = set()
+    for fn in filenames or ():
+        m = _WHEEL_BUILD_TAG_RE.match(fn)
+        if m:
+            tags.add(m.group("build"))
+    if not tags:
+        return ""
+    return "#builds=" + ",".join(sorted(tags))
```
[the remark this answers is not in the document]
ryanpetrello: PEP 427 build tags are alphanumeric values that start with a digit e.g.,

```
1
2
2a
2beta
3rc1
```

...so this isn't necessary.

pulp_python/app/pypi/feeds.py
```diff
@@ -91,14 +95,30 @@ def iter_projects(content, repo_ver, limit=PACKAGES_LIMIT):
     )
 
 
-def _item_dict(title, link, description, author_email, pubdate):
+def _build_tag_fragment(filenames):
+    """Extract sorted distinct build tags from wheel filenames for GUID stability.
+
+    Returns a fragment like ``#builds=1,2`` when build tags are present,
```
gerrod3, line 101: Use single ticks.

pulp_python/tests/unit/test_feeds.py
```diff
@@ -0,0 +1,67 @@
+import re
+
+import pytest
+
+# Duplicated here to avoid importing feeds.py, which pulls in Django/DRF and
+# requires a configured Django settings module that the unit test runner lacks.
```
gerrod3, line 6: If you add pytest-django to unittest_requirements.txt then you could import this constant with no issue.

CHANGES/1381.bugfix
```diff
@@ -0,0 +1 @@
+RSS feed GUIDs are now stable when additional platform wheels arrive for the same release. GUIDs only change when a new PEP 427 build tag is introduced, preventing feed readers from showing duplicate entries for multi-architecture builds.
```
gerrod3, line 1: I know this feature is one you requested, but I wonder if we are deviating from PyPI's behavior. Can you show me an example of a package on PyPI that updates their entries based on build-tags?

gerrod3 (changes requested):

CHANGES/1381.bugfix
ryanpetrello: That's a pretty good point 🤔 - we may be coloring this RSS implementation with our desired behavior, and not with parity in terms of what PyPI actually does.

ryanpetrello: Yea, @mprpic, we may actually be at an impasse here.  Looking at what PyPI actually exposes in its RSS feed, it doesn't provide a `<guid>` element, which means it just falls back to the `<link>`.

PyPI's link is https://pypi.org/project/{name}/{version}/ -- one entry per (name, version), no timestamp, no build tag.

So if we're modeling this after the way that PyPI behaves, the link is the GUID; adding more wheels to an existing release doesn't create a new RSS entry.

So even though the original limitation you discovered doesn't work the way we expected it to, it matches what PyPI does (and I understand why Pulp maintainers would want to mirror the PyPI implementation).

gerrod3: Looking at an example it seems that PyPI doesn't even use GUID: https://pypi.org/rss/project/tensorflow/releases.xml

On the other hand maybe this is a limitation of PyPI and their RSS feeds would be more useful with this.

gerrod3 (approved):

